PR #12, opened 2021-10-31
Feat/add blocking request

### LINKED ISSUE #7: Feature Request: option to block request until container loads
Thank you for this cool idea! I am hoping to use it to reduce the load on a server that has a few rarely-used containers.

It would be great to have a feature that simply keeps the request open until the container is up and running, without a loading screen.

A great use case for this is a simple container I run that serves a single json file (a .wellknown file). If I put the ondemand plugin on this, it would break clients as they don't know how to parse the loading screen. However, if it simply blocked until the container is up and running, it would work.

An additional parameter for "wait" time after a container starts may be necessary if it takes some time for the container to be ready to serve an image.

acouvreur:
Keeps the request hanging until the container is up and running, without a loading screen.

Add a serving strategy for further changes. Each strategy have its own configuration and handles the request.

Closes #7


ondemand.go
```diff
@@ -3,40 +3,33 @@ package traefik_ondemand_plugin
...
-	Timeout: time.Second * 2,
-}
-
 // Config the plugin configuration
 type Config struct {
 	Name       string `yaml:"name"`
 	ServiceUrl string `yaml:"serviceurl"`
 	Timeout    string `yaml:"timeout"`
+	WaitUi     bool   `yaml:"wait-ui"`
+	BlockDelay string `yaml:"block-delay"`
```
tomasharkema, line 18: I've tested this branch, looks like the hyphen is not allowed in some way. I can confirm the following works:

```go
	WaitUi     bool   `yaml:"waitui"`
	BlockDelay string `yaml:"blockdelay"
```
with config
```yaml
          waitui: false
          blockingdelay: 20s
```

Otherwise WaitUi will remain true based on the default config from `CreateConfig`

pkg/strategy/blocking_strategy.go
```diff
@@ -0,0 +1,44 @@
...
+	Next       http.Handler
+	Timeout    time.Duration
+	BlockDelay time.Duration
+}
+
+// ServeHTTP retrieve the service status
+func (e *BlockingStrategy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+
+	// TODO: Wait for 
+	for start := time.Now(); time.Since(start) < e.BlockDelay; {
```
tomasharkema, line 24: I tested it some more! It would appear this code will hammer the on demand service, making traefik and the service hang. I tried to come up with some back-off system, but since I'm no where proficient in Go, I only could come up with adding `time.Sleep(1 * time.Second)` at the end of the for loop. No idea if that has implications for blocking the thread and maybe it blocks other behavior. Otherwise a non blocking backoff system would be better to schedule retrying and keep the thread free for other work. Or is this not something to worry about?

acouvreur: It indeed needs to wait before retrying, otherwise it will hammer the service as you described.
A simple wait should not block any other request.
Each request is handled by a goroutine, thus not blocking any other work from the app.

ondemand.go
acouvreur: It should work with hyphens, so I should find out why it doesn't

Issue #6, opened 2021-09-28
custom loading and error page


acouvreur:
Allow custom loading and error page by giving a file path

PR #15, opened 2021-12-08
Custom loading page

sourgrasses:
I implemented the custom loading page feature mentioned in issue #6 and added a little documentation for it. I think the implementation wound up being pretty straight-forward, but I'm happy to change something if you think it could be written better or in a more idiomatic way.

### DIFF
--- a/pkg/pages/error.go
+++ b/pkg/pages/error.go
@@ -175,15 +175,26 @@ type ErrorData struct {
 	Error string
 }
 
-func GetErrorPage(name string, e string) string {
-	tpl, err := template.New("error").Parse(errorPage)
+func GetErrorPage(path, name string, e string) string {
+	var tpl *template.Template
+	var err error
+	if template_path != "" {
+		tpl, err = template.New(path.Base(template_path)).ParseFiles(template_path)
+	} else {
+		tpl, err = template.New("loading").Parse(loadingPage)
+	}
 	if err != nil {
 		return err.Error()
 	}
+
 	b := bytes.Buffer{}
-	tpl.Execute(&b, ErrorData{
+	err = tpl.Execute(&b, ErrorData{
 		Name:  name,
 		Error: e,
 	})
+	if err != nil {
+		return err.Error()
+	}
+
 	return b.String()
 }
--- a/pkg/pages/loading.go
+++ b/pkg/pages/loading.go
@@ -2,6 +2,7 @@ package pages
 
 import (
 	"bytes"
+	"path"
 
 	"fmt"
 	"html/template"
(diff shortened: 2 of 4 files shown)


acouvreur:
Thanks for this pull request !
I will give you a review, but it seems promising already :)

pkg/pages/error.go
```diff
@@ -175,15 +175,26 @@ type ErrorData struct {
 	Error string
 }
 
-func GetErrorPage(name string, e string) string {
-	tpl, err := template.New("error").Parse(errorPage)
+func GetErrorPage(path, name string, e string) string {
```
acouvreur, line 178: I think it should be should be 
```diff
- func GetErrorPage(path, name string, e string) string {
+ func GetErrorPage(template_path string, name string, e string) string {
```

pkg/pages/error.go
```diff
@@ -175,15 +175,26 @@ type ErrorData struct {
...
+	var tpl *template.Template
+	var err error
+	if template_path != "" {
+		tpl, err = template.New(path.Base(template_path)).ParseFiles(template_path)
```
acouvreur, line 182: `path` is missing from the import list

pkg/pages/loading.go
```diff
@@ -251,16 +252,27 @@ type LoadingData struct {
 	Timeout string
 }
 
-func GetLoadingPage(name string, timeout time.Duration) string {
-	tpl, err := template.New("loading").Parse(loadingPage)
+func GetLoadingPage(template_path, name string, timeout time.Duration) string {
```
acouvreur, line 255: I think you should specify template_path type

```diff
-  func GetLoadingPage(template_path, name string, timeout time.Duration) string {
+  func GetLoadingPage(template_path string, name string, timeout time.Duration) string {
```

acouvreur (changes requested):

acouvreur:
We should also specify in the documentation a link to the default templates /pkg/pages/error.html and /pkg/pages/loading.html and their respective data structure forwarded to the template.
So developers who wants to create their own page know what data they have access to.

pkg/pages/error.go
sourgrasses: Oh whoops, yeah, I must've managed to undo a rename partially just before I pushed that.

sourgrasses:
Alright, I made those fixes and changes and updated the readme to link to the templates. I noticed also that those html files don't quite match the templates that actually get rendered. Should I go ahead and update those to match?

acouvreur:
I wanted to use https://pkg.go.dev/embed instead of duplicating the defaults as strings in the files.
But as far as I tested, I cannot use it with the way Yaegi includes the plugin.

sourgrasses:
Ah yeah, Yaegi does make things a little more fiddly there.

acouvreur (approved):

acouvreur:
LGTM!

PR #20, opened 2022-02-11
feat: Add support for names to scale down multiple containers

Nastaliss:
[Original issue ](https://github.com/acouvreur/traefik-ondemand-plugin/issues/19)

Using `Name` is still allowed, but should be removed in the next major bump. Using Names as a one-element array does the same thing. 

I looked into adding the services list to the default loading page. But it required a rework on the variables passed to the strategies. I'll work towards this in the future. 

I factorized the test cases so that it's easier to add more tests / test cases.


acouvreur:
Thanks for the pull request, I'm quit busy right now, but it seems promising!
I'll give it a better look soon.

And I think we'll merge this feature into beta channel first.

Nastaliss:
Alright awesome ! 
I'm testing it further this week.
I'll also change the base to `beta` .

PR #65, opened 2022-10-28
Feature/stabilization

acouvreur:
This stabilize the next development interation and is a total architecture rewrite.

It also brings themes and all the multiple instance handling is done backend side


acouvreur:
TODO:
- Docs
- Response body for instances

PR #80, opened 2022-11-04
Feature/add-docker-event-listener

acouvreur:
When a container is shutdown manually while it's registered as `ready` in Sablier it will remove it from the store.
Meaning externally events are now handled for docker.

WIP Commit

### DIFF
--- a/app/providers/docker_classic.go
+++ b/app/providers/docker_classic.go
@@ -2,16 +2,19 @@ package providers
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
 
 	"github.com/acouvreur/sablier/app/instance"
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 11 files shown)


app/providers/docker_classic.go
```diff
@@ -103,3 +106,32 @@ func (provider *DockerClassicProvider) GetState(name string) (instance.State, er
...
+	go func() {
+		for {
+			select {
+			case msg := <-msgs:
+				// Send the container that has died to the channel
+				instance <- msg.From
+			case err := <-errs:
+				if errors.Is(err, io.EOF) {
+					log.Debug("provider event stream closed")
+					close(instance)
```
acouvreur, line 128: This doesnt follow "The Channel Closing Principle"

One general principle of using Go channels is don't close a channel from the receiver side and don't close a channel if the channel has multiple concurrent senders. In other words, we should only close a channel in a sender goroutine if the sender is the only sender of the channel.

See https://go101.org/article/channel-closing.html#:~:text=The%20Channel%20Closing%20Principle,only%20sender%20of%20the%20channel.

PR #87, opened 2022-11-07
Feature/add-kubernetes-event-listener

acouvreur:
If a deployment or a statefulset is scaled to 0, then it is removed from the store.

In order for this to work, you need the `list` and `watch` verbs in your ClusterRole


acouvreur:
Add tests and documentation

PR #122, opened 2022-12-03
feat(plugins): add nginx plugin

### LINKED ISSUE #118: Add `nginx` reverse proxy integration
**Describe the reverse proxy you'd like**
Nginx reverse proxy

**Does the reverse proxy provides middleware/module/extension support**
Yes through 3rd party modules.

- https://www.nginx.com/resources/wiki/modules/
- http://nginx.org/en/docs/dev/development_guide.html

> Common Pitfalls
> Writing a C module
> 
> The most common pitfall is an attempt to write a full-fledged C module when it can be avoided. In most cases your task can be accomplished by creating a proper configuration. If writing a module is inevitable, try to make it as small and simple as possible. For example, a module can only export some [variables](http://nginx.org/en/docs/dev/development_guide.html#http_variables).
> 
> Before starting a module, consider the following questions:
> 
>    Is it possible to implement a desired feature using already [available modules](http://nginx.org/en/docs/index.html)?
>    Is it possible to solve an issue using built-in scripting languages, such as [Perl](http://nginx.org/en/docs/http/ngx_http_perl_module.html) or [njs](http://nginx.org/en/docs/njs/index.html)?

- http://nginx.org/en/docs/njs/index.html

**Additional context**

We need to be able to create a module that can do the following op

So be able to intercept a packet, and either return custom data or proxy pass depending on the subrequest result.

acouvreur:
Work in progress

TODO:

- [ ] Add e2e tests
- [ ] Add development documentation
- [ ] Check whether this configuration has enough flexibility

closes #118

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -81,3 +81,38 @@ jobs:
 
       - name: Test ${{ matrix.provider }}
         run: cd plugins/traefik/e2e && bash ./${{ matrix.provider }}.sh
+
+  nginx_e2e:
+    name: Run Sablier E2E tests for Nginx NJS module with Sablier
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      fail-fast: false
+      matrix:
+        provider: [docker_classic, docker_swarm, kubernetes]
+    steps:
+      - name: Set up Go 1.18
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.18
+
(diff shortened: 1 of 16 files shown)


acouvreur:
Should I release the javascript file? Or also a precompiled nginx ? I'm not so sure.

Prebaked nginx for sablier makes sense as I can guarantee nginx compatibility, but some people might not like it. So maybe both options altogether.

PR #134, opened 2023-03-10
Feature/add filter by labels

acouvreur:
You can tag a container with the following:

```yaml
  whoami:
    image: containous/whoami:v1.5.0
    labels:
      - sablier.enable=true
      # Optional, defaults to "default"
      - sablier.group=whoami
```

And then use the Traefik middleware with the following 
```yaml
      # Blocking Middleware
      # This is the relevant change, here we specify group instead of "names"
      - traefik.http.middlewares.blocking.plugin.sablier.group=sablier
      - traefik.http.middlewares.blocking.plugin.sablier.sablierUrl=http://sablier:10000
      - traefik.http.middlewares.blocking.plugin.sablier.sessionDuration=1m
      - traefik.http.middlewares.blocking.plugin.sablier.blocking.timeout=30s
```

The API call allows now either "names" or "group" as a query parameter.

How to try it yourself when developing:

1. git clone the repo
2. `docker run -d --name=whoami --label "sablier.enable=true" whoami`
3. docker stop whoami
4. go run main.go --logging.level=debug start
5. curl http://localhost:10000/api/strategies/blocing?group=default&session_duration=1m

You should now see whoami started.

The same steps can be reproduced for swarm.


tam481:
Does this mean that all you would need to do is add the label to the container and Sablier would then stop and start it automatically as and when required? I'm confused because I thought you needed the container info to be in the dynamic config file because Traefik will not recognise the container when it's shutdown!

acouvreur:
> Does this mean that all you would need to do is add the label to the container and Sablier would then stop and start it automatically as and when required? I'm confused because I thought you needed the container info to be in the dynamic config file because Traefik will not recognise the container when it's shutdown!

So this mechanism is not related to Traefik exactly.
It's just an alternative to specify a group of container to handle per middleware.

Instead of specifying each individual name for your provider (which can change across time...) you can specify a group name which gets fetched periodically in the background by the provider.

tam481:
So how does it work? Do I add the label and the middleware like this:
      - "sablier.enable=true"
      - traefik.http.middlewares.dynamic.plugin.sablier.names=sablier_dockerContainer
      - traefik.http.middlewares.dynamic.plugin.sablier.sablierUrl=http://sablier:10000
      - traefik.http.middlewares.dynamic.plugin.sablier.sessionDuration=1m
      - traefik.http.middlewares.dynamic.plugin.sablier.dynamic.theme=hacker-terminal

acouvreur:
> So how does it work? Do I add the label and the middleware like this: - "sablier.enable=true" - traefik.http.middlewares.dynamic.plugin.sablier.names=sablier_dockerContainer - traefik.http.middlewares.dynamic.plugin.sablier.sablierUrl=http://sablier:10000 - traefik.http.middlewares.dynamic.plugin.sablier.sessionDuration=1m - traefik.http.middlewares.dynamic.plugin.sablier.dynamic.theme=hacker-terminal

As stated in my example, you have a new key on your configuration. It's not a breaking change so you can still use the `names` parameter, but a new one is available with the key `group`. See:

```
      # Blocking Middleware
      # This is the relevant change, here we specify group instead of "names"
      - traefik.http.middlewares.blocking.plugin.sablier.group=sablier
      - traefik.http.middlewares.blocking.plugin.sablier.sablierUrl=http://sablier:10000
      - traefik.http.middlewares.blocking.plugin.sablier.sessionDuration=1m
      - traefik.http.middlewares.blocking.plugin.sablier.blocking.timeout=30s
```

Please correct me if I'm wrong, but from my understanding the labels are attached to the container you actually want to shut down on demand right?
So, in your case it'd be changed to

```
- sablier.enable=true
- traefik.http.middlewares.dynamic.plugin.sablier.group=default
- traefik.http.middlewares.dynamic.plugin.sablier.sablierUrl=http://sablier:10000/
- traefik.http.middlewares.dynamic.plugin.sablier.sessionDuration=1m
- traefik.http.middlewares.dynamic.plugin.sablier.dynamic.theme=hacker-terminal
```

Note the `default` value. This is because this is the default group for your containers when the group name is not set.

If you want a custom group, you'd do the following:

```diff
 - sablier.enable=true
+ - sablier.group=mygroup
- - traefik.http.middlewares.dynamic.plugin.sablier.group=default
+ - traefik.http.middlewares.dynamic.plugin.sablier.group=mygroup
 - traefik.http.middlewares.dynamic.plugin.sablier.sablierUrl=http://sablier:10000/
 - traefik.http.middlewares.dynamic.plugin.sablier.sessionDuration=1m
 - traefik.http.middlewares.dynamic.plugin.sablier.dynamic.theme=hacker-terminal
```

Does it make sense?

tam481:
Thank you.

I added that to one of my containers but nothing happens. Am I missing something?
In the Traefik dashboard I cannot see middleware associated with the container. However, the two other containers that I configured in the dynamic config file are showing their Sablier middleware. What am I missing?

p.s. I'm running plugin v1.3 and the latest beta release of the container.

acouvreur:
> Thank you.
> 
> I added that to one of my containers but nothing happens. Am I missing something? In the Traefik dashboard I cannot see middleware associated with the container. However, the two other containers that I configured in the dynamic config file are showing their Sablier middleware. What am I missing?
> 
> p.s. I'm running plugin v1.3 and the latest beta release of the container.

Hi @tam481,

You should always use the same version for the plugin and the docker image.

There was an issue publishing the beta version on docker, I have to figure that out.

If you use the beta version of the docker image you will have group scanning capabilities.

However, if you do not use the beta version from the middleware, it won't be able to send the correct request stating the group.

As the plugin is only registered for tags, please consider using `v1.4.0-beta.3` 
image

The docker image version `beta` should point to the same version of the ode. However, as this tag gets overwrite on every beta release, make sure to use `docker pull acouvreur/sablier:beta`.

I'll fix the release process on tags for docker image, it should be less confusing in the future.

tam481:
I didn't realise that the plugin follows the main application. Thank you for that.

Does this look correct for the plugin:

experimental:
  plugins:
    sablier:
      moduleName: "github.com/acouvreur/sablier"
      version: "v1.4.0-beta.3"

acouvreur:
> I didn't realise that the plugin follows the main application. Thank you for that.
> 
> Does this look correct for the plugin:
> 
> experimental: plugins: sablier: moduleName: "github.com/acouvreur/sablier" version: "v1.4.0-beta.3"

Yes, I think it looks good.

Provide the overall configuration so I can be sure that you configured everything right (traefik + sablier + service to set on demand)

tam481:
Here's my container config:

`  obooks:
    image: evanbuss/openbooks:latest
   # restart: unless-stopped
    container_name: obooks
    hostname: obooks
    volumes:
      - "$PWD/../conf/obooks:/evanbuss/books"
      - "/mnt/share/books:/books"
    networks:
      - backend
    command: 
      --persist
      --name Star-node
    labels:
      - "traefik.enable=true"
      - "traefik.docker.network=backend"
      ## HTTP Routers/Service
      - "traefik.http.routers.books-rtr.entrypoints=web-ep"
      - "traefik.http.routers.books-rtr.rule=Host(`books.${EXT_DOMAIN}`)"
      - "traefik.http.routers.books-rtr.tls=true"
      - "traefik.http.routers.books-rtr.tls.certresolver=mydefault"
      ## HTTP Services
      - "traefik.http.routers.books-rtr.service=books-svc"
      - "traefik.http.services.books-svc.loadbalancer.server.port=80"
      ## Sablier
      - "sablier.enable=true"
      ## Middleware
      - traefik.http.middlewares.dynamic.plugin.sablier.group=default
      - traefik.http.middlewares.sablier_books.plugin.sablier.sablierUrl=http://sablier:10000
      - traefik.http.middlewares.sablier_books.plugin.sablier.sessionDuration=1m
      - traefik.http.middlewares.sablier_books.plugin.sablier.dynamic.theme=hacker-terminal`

tam481:
Had to edit it a few times to get it to display correctly. Sorry about that.
It doesn't seem to work. I would expect Sablier to shutdown the container after 1 minute but nothing happens. In the Traefik dashboard, I cannot see anything under "middleware" for the container

acouvreur:
> Had to edit it a few times to get it to display correctly. Sorry about that. It doesn't seem to work. I would expect Sablier to shutdown the container after 1 minute but nothing happens. In the Traefik dashboard, I cannot see anything under "middleware" for the container

From your input I can see that you are using docker provider, not swarm or kubernetes.

Traefik has some limitation on this part. See https://github.com/acouvreur/sablier/tree/main/plugins/traefik#traefik-with-docker-classic

Basically, all containers that are stopped are removed from the pool, thus all metadata attatched to them as labels.

They build workaround for swarm and kubernetes but not for docker classic.

You'll have to use a dynamic config file for traefik metadata.
The `sablier.enable` label can be kept on your container as I also scan for stopped containers.

PR #137, opened 2023-03-26
fix docker-compose to bring in line with quick-start

Sam-R:
Minor fixes to the `docker-compose.yml` file based on the quick-start instructions in `README.md`.

- update docker image to use `acouvreur/sablier:1.3.0` as per dockerhub's tag to solve `ERROR: manifest for acouvreur/sablier:v1.3.0 not found: manifest unknown: manifest unknown` message on `docker-compose up -d`.
- update container name to use underscores (default on Docker 23.0.1 on Ubuntu 22.04) to solve `An error occured: Error: No such container: sablier-whoami-1` error when calling the dynamic URL in a web browser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ sablier --help
 # or
 
 docker run acouvreur/sablier[:version] --help
-# ex: docker run acouvreur/sablier:v1.3.0 --help
+# ex: docker run acouvreur/sablier:1.3.0 --help
 ```
 
 All arguments can be used in the form of the config file such as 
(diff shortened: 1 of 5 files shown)


acouvreur:
Oh you're right about the tag!
I'll check for the containers names.

acouvreur:
I superseded your PR changes for the version with https://github.com/acouvreur/sablier/pull/138.

Can you rebase on it once merged ?

And we should probably use container names instead of the auto assigned docker-compose ones.

What do you think?

Sam-R:
Sure, I've rebased and the PR is now up to date with main.

I tend to avoid specific container names in my environments because it impacts dockers ability to scale. For the quick-start it might be okay but I wouldn't normally run named containers in pre-prod/prod.

Are you seeing different docker container name formats? I wasn't sure if that'd be the case which is why I listed my docker version and OS version.

PR #146, opened 2023-04-07
fix: dynamic configuration of plugin

tandy-1000:


acouvreur:
Thanks for you change! It is indeed an mistake in the documentation!

PR #166, opened 2023-08-07
Update sessions.go

mscreations:
Fix incorrect spelling for Interval

Not sure if this would cause an error or not. Noticed it while reviewing code.


acouvreur:
Thanks @mscreations :)

PR #167, opened 2023-08-08
feat(api) Added Status API

ab623:
I have recently picked up learning Go and trying to learn to contribute back into open source. And thought that Sablier would be a good project to get involved in.

I have been putting in the early stages of a status page, which intends to show the status of the services Sablier manages, along with useful information about those services, mainly, the status, expiration time, and remaining time.

I had to make quite a few updates across the solution. I have documented the rationale below

* Gin - Updates to include status page. The response payload is designed in a way so that it can be extended with global configs. 
* KV store - This required an update of a new method to be able to get a value along with expiration information 
* Instance - Updated structs to store additional state data
* Providers - Updated providers with a new function to get the list of managed containers, using label enable = true
* Session Manager - Few methods to get the service status

Output:
Calling `/api/status` :

```json
{
  "managed_services": [
    {
      "name": "whoami-secondary",
      "currentReplicas": 1,
      "desiredReplicas": 1,
      "status": "ready",
      "expires_on": "0001-01-01T00:00:00Z",
      "expires_in": 0
    },
    {
      "name": "whoami-primary",
      "currentReplicas": 0,
      "desiredReplicas": 1,
      "status": "unrecoverable",
      "message": "ready"
      "expires_on": "2023-08-08T11:57:59.3412899Z",
      "expires_in": 260
    }
  ]
}
```

Potential non-blocking issues:
* The status could be refined a bit better depending on if the container is up or down
* If a service hasn't been hit for the first time, it shows as ready with no expiry (see whoami-secondary). This has been documented in https://github.com/acouvreur/sablier/issues/153, and could be a quick fix on startup.

I haven't used Go before, nor any of the libraries such as Gin, so some good feedback and advice would be helpful.  I have tired to stick close to the  format in the project.

### DIFF
--- a/app/http/routes/status.go
+++ b/app/http/routes/status.go
@@ -1,0 +1,52 @@
+package routes
+
+import (
+	"net/http"
+
+	"github.com/acouvreur/sablier/app/instance"
+	"github.com/acouvreur/sablier/app/sessions"
+	"github.com/acouvreur/sablier/config"
+	"github.com/gin-gonic/gin"
+)
+
+type Status struct {
+	SessionsManager sessions.Manager
+	StrategyConfig  config.Strategy
+	SessionsConfig  config.Sessions
+}
+
+func NewStatus(sessionsManager sessions.Manager, strategyConf config.Strategy, sessionsConf config.Sessions) *Status {
+
+	status := &Status{
+		SessionsManager: sessionsManager,
+		StrategyConfig:  strategyConf,
+		SessionsConfig:  sessionsConf,
+	}
+
+	return status
+}
+
+type StatusResponse struct {
+	ManagedServices []instance.State `json:"managed_services"`
+}
+
+func (s *Status) ServeStatus(c *gin.Context) {
+	containers, _ := s.SessionsManager.GetManagedContainers()
+
+	var managedServices []instance.State
+	for _, name := range containers {
+
+		state := s.SessionsManager.GetContainerStatus(name)
+
+		if state.Error != nil {
+			c.AbortWithStatus(500)
+		}
+
+		managedServices = append(managedServices, *state.Instance)
+	}
+
+	output := StatusResponse{
+		ManagedServices: managedServices,
+	}
+	c.JSON(http.StatusOK, output)
+}
--- a/app/instance/instance.go
+++ b/app/instance/instance.go
@@ -1,17 +1,23 @@
 package instance
 
-import log "github.com/sirupsen/logrus"
+import (
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
 
 var Ready = "ready"
 var NotReady = "not-ready"
 var Unrecoverable = "unrecoverable"
 
 type State struct {
-	Name            string `json:"name"`
-	CurrentReplicas int    `json:"currentReplicas"`
-	DesiredReplicas int    `json:"desiredReplicas"`
-	Status          string `json:"status"`
-	Message         string `json:"message,omitempty"`
+	Name            string        `json:"name"`
+	CurrentReplicas int           `json:"currentReplicas"`
+	DesiredReplicas int           `json:"desiredReplicas"`
+	Status          string        `json:"status"`
+	Message         string        `json:"message,omitempty"`
+	ExpiresAt       time.Time     `json:"expires_on"`
+	ExpiresAfter    time.Duration `json:"expires_in"`
 }
 
 func (instance State) IsReady() bool {
--- a/app/instance/instance.go
+++ b/app/instance/instance.go
@@ -17,7 +17,7 @@ type State struct {
 	Status          string        `json:"status"`
 	Message         string        `json:"message,omitempty"`
 	ExpiresAt       time.Time     `json:"expires_on"`
-	ExpiresAfter    time.Duration `json:"expires_in"`
+	ExpiresIn       time.Duration `json:"expires_in"`
 }
 
 func (instance State) IsReady() bool {
--- a/app/sessions/sessions_manager.go
+++ b/app/sessions/sessions_manager.go
@@ -277,7 +277,7 @@ func (s *SessionsManager) RequestReadySessionGroup(ctx context.Context, group st
 
 func (s *SessionsManager) GetManagedContainers() ([]string, error) {
 
-	output, err := s.provider.GetMangedContainers()
+	output, err := s.provider.GetManagedContainers()
 
 	if err != nil {
 		return nil, fmt.Errorf("Error getting list of managed containers")
--- a/app/sessions/sessions_manager.go
+++ b/app/sessions/sessions_manager.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/acouvreur/sablier/app/instance"
 	"github.com/acouvreur/sablier/app/providers"
+	"github.com/acouvreur/sablier/config"
 	"github.com/acouvreur/sablier/pkg/tinykv"
 	log "github.com/sirupsen/logrus"
 )
@@ -34,12 +35,13 @@ type SessionsManager struct {
 	ctx    context.Context
 	cancel context.CancelFunc
 
+	config   config.Config
 	store    tinykv.KV[instance.State]
 	provider providers.Provider
 	groups   map[string][]string
 }
 
-func NewSessionsManager(store tinykv.KV[instance.State], provider providers.Provider) Manager {
+func NewSessionsManager(store tinykv.KV[instance.State], provider providers.Provider, conf config.Config) Manager {
 	ctx, cancel := context.WithCancel(context.Background())
 
 	groups, err := provider.GetGroups()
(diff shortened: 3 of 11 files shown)


ab623:
Thanks @acouvreur. Additionally I just pushed a commit to fix https://github.com/acouvreur/sablier/issues/153. I did it on this branch as it requires the functionality in the initial commits.

This works by checking all managed containers on startup, and if the containers are running already, it just applies the default timeout to them. If a containers is already stopped it just skips it.

This way the containers can start first, and then Sabiler can be started after and the containers that it manages will be brought down after the default timeout.

acouvreur:
So for the status page we have a few things to consider:
- Instances that have been started by name
- Instances that have been started by group (auto discovery)

If it is started by name, we will only see the status when it's up, because as soon as it's down, the instance will be out of the KV store.

For instances started with the group name, then you can already use the group values to build the status page.
So I think you do not need to add anything to the providers.

What do you think about that?

ab623:
My implementation already accounts for that and is independent of the calling method.

It looks for the enable label and uses that to track all monitored containers.

Then if it is up, it exists in the KV and pulls the expiration data, if not then the service is down, but the container name is still shown in the output, just with a exited status.

app/instance/instance.go
```diff
@@ -1,17 +1,23 @@
...
-	CurrentReplicas int    `json:"currentReplicas"`
-	DesiredReplicas int    `json:"desiredReplicas"`
-	Status          string `json:"status"`
-	Message         string `json:"message,omitempty"`
+	Name            string        `json:"name"`
+	CurrentReplicas int           `json:"currentReplicas"`
+	DesiredReplicas int           `json:"desiredReplicas"`
+	Status          string        `json:"status"`
+	Message         string        `json:"message,omitempty"`
+	ExpiresAt       time.Time     `json:"expires_on"`
```
acouvreur, line 19: The property name is ExpiresAt but the json is expires_on
We should use one exclusively

app/http/routes/status.go
```diff
@@ -0,0 +1,52 @@
...
+
+	return status
+}
+
+type StatusResponse struct {
+	ManagedServices []instance.State `json:"managed_services"`
+}
+
+func (s *Status) ServeStatus(c *gin.Context) {
+	containers, _ := s.SessionsManager.GetManagedContainers()
```
acouvreur, line 34: It's not always containers, it might be swarm services, kubernetes deployments, etc.
We should use `instances` instead.

app/sessions/sessions_manager.go
```diff
@@ -69,6 +75,30 @@ func (sm *SessionsManager) initWatchers() {
...
 }
 
+func (sm *SessionsManager) initContainerManagementStart() {
+	output, err := sm.provider.GetManagedContainers()
+
+	if err != nil {
+		log.Debugf("Could not get list of managed containers to apply default")
+	}
+
+	var startContainers []string
```
acouvreur, line 85: It's not necessarily containers, we should use `instances`

app/sessions/sessions_manager.go
```diff
@@ -69,6 +75,30 @@ func (sm *SessionsManager) initWatchers() {
 	go sm.consumeInstanceStopped(instanceStopped)
 }
 
+func (sm *SessionsManager) initContainerManagementStart() {
+	output, err := sm.provider.GetManagedContainers()
```
acouvreur, line 79: Instead of `GetManagedContainers`, you should use `sm.groups` which is already populated with the discovered instances

app/sessions/sessions_manager.go
```diff
@@ -69,6 +75,30 @@ func (sm *SessionsManager) initWatchers() {
...
+		if err != nil {
+			log.Debugf("Could not get state for %s", name)
+		}
+
+		if instance.IsReady() {
+			startContainers = append(startContainers, name)
+		}
+	}
+
+	sm.RequestSession(startContainers, sm.config.Sessions.DefaultDuration)
```
acouvreur, line 99: I'm not really sure about this. I believe the behavior is that we want to **stop** the instances started we found which are **not** in the KV.

app/sessions/sessions_manager.go
```diff
@@ -273,6 +303,38 @@ func (s *SessionsManager) RequestReadySessionGroup(ctx context.Context, group st
 	return s.RequestReadySession(ctx, names, duration, timeout)
 }
 
+func (s *SessionsManager) GetManagedContainers() ([]string, error) {
```
acouvreur, line 306: Not necessarily containers, use instance instead

app/sessions/sessions_manager.go
```diff
@@ -273,6 +303,38 @@ func (s *SessionsManager) RequestReadySessionGroup(ctx context.Context, group st
...
+	output, err := s.provider.GetManagedContainers()
+
+	if err != nil {
+		return nil, fmt.Errorf("Error getting list of managed containers")
+	}
+
+	return output, nil
+}
+
+func (s *SessionsManager) GetContainerStatus(name string) *InstanceState {
```
acouvreur, line 317: Not necessarily containers, use instance instead

pkg/tinykv/tinykv.go
```diff
@@ -129,6 +136,41 @@ func (kv *store[T]) Get(k string) (T, bool) {
...
 
+func (kv *store[T]) GetWithTimeout(k string) (EntityWithTimeout, bool) {
+	kv.mx.Lock()
+	defer kv.mx.Unlock()
+
+	e, ok := kv.kv[k]
+	if !ok {
+		return EntityWithTimeout{}, ok
+	}
+	if e.expired() {
```
acouvreur, line 147: I don't think this method should do any check on the timeouts and should not delete from the KV.

I think the method should be `GetExpiration` for a key.

acouvreur (changes requested):
Overall I agree with your changes, but I think we need to generify the usage to instance instead of container.

Also, I believe the startup feature is not as expected. We should review it again. And it should also be enabled through a configuration flag.

app/instance/instance.go
ab623: Any preference on this?

app/http/routes/status.go
ab623: Makes sense here. I only have docker experience. I will look into using instances.

app/sessions/sessions_manager.go
ab623: Okay using sm.groups does make sense and simplified things. In that case we don't need to update the interface, and I can remove the "GetManagedContainers".

I think I will need to flatten a copy of the sm.groups, as this is a map with the group name as the key. And I believe the status API should be flat with the group as a property of the instance, rather than being grouped by groupname (which the API consumer can do on the client side)

app/sessions/sessions_manager.go
ab623: I disagree with this partially. I believe that if I have an already running containers, and I restart sablier, it would bring my instances down instantly. I prefer having it this way, but I also see your argument for instantly bringing them down.

Maybe we can create a setting that is like STARTUP_DEFAULT_EXPIRATION_DURATION which can be set to "0m" for instant or whatever the user sets. If the setting it blank it defaults to the SESSIONS_DEFAULT_DURATION.

Would that work for you?

pkg/tinykv/tinykv.go
ab623: I don't fully agree with this. If I wanted to get a value and an expiration, under your suggestion I would need to call 2 separate functions, which COULD cause an issue whereby, I manage to retrieve the value, but in the time between calling `Get` and `GetExpiration` the key expired, it would cause an error because the Key would have expired. 

Keeping it as 1 call to get the value and expiration as a single struct makes more sense in my opinion. Why do you think we shouldn't be checking on timeout or deleting from KV like `Get` does?

ab623:
> Overall I agree with your changes, but I think we need to generify the usage to instance instead of container.
> 
> Also, I believe the startup feature is not as expected. We should review it again. And it should also be enabled through a configuration flag.

Thanks for the feedback. Glad my first foray into the Go world and this project was somewhat successful.

I will refactor this to use the instances and not containers as per your suggestions and review the startup feature. I have left a few comments on both. Would be good to get your thoughts on them.

acouvreur:
> > Overall I agree with your changes, but I think we need to generify the usage to instance instead of container.
> > Also, I believe the startup feature is not as expected. We should review it again. And it should also be enabled through a configuration flag.
> 
> Thanks for the feedback. Glad my first foray into the Go world and this project was somewhat successful.
> 
> I will refactor this to use the instances and not containers as per your suggestions and review the startup feature. I have left a few comments on both. Would be good to get your thoughts on them.

Take your time for the refactor by the way, if you need some help feel free to ask me for it.

Thanks again for your contribution by the way :)

ab623:
Thanks. I'm travelling for a few weeks. So won't have time to work on this, but will pick it up when I'm back.

Appreciate the support.

I left a few comments on the review can you take a look and let me know your thoughts.

ab623:
Hey @acouvreur. I'm back from some holidays. Should I continue to work on this, or wait until the rewrite? Or are you building this in as part of the rewrite 😄

ab623:
Alright. Good to know! I'll close this PR. I'll give the refactor branch a go also.

PR #178, opened 2023-09-13
fix(providers/kubernetes): Added QPS and Burst tweaks for client-side throttling

valexz:
Hello

The reasoning behind this PR is to enable the adjustment of QPS and Burst configuration parameters in the client-go library. 

This will help avoid constant throttling issues when dealing with over 100 K8S deployments. The default values are insufficient for such a large number of deployments.

How client-side throttling works:
      // NewTokenBucketRateLimiter creates a rate limiter which implements a token bucket approach.
      // The rate limiter allows bursts of up to 'burst' to exceed the QPS, while still maintaining a
      // smoothed qps rate of 'qps'.
      // The bucket is initially filled with 'burst' tokens, and refills at a rate of 'qps'.
      // The maximum number of tokens in the bucket is capped at 'burst'.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     image: traefik:2.9.1
     command:
       - --experimental.plugins.sablier.moduleName=github.com/acouvreur/sablier
-      - --experimental.plugins.sablier.version=v1.4.0-beta.5
+      - --experimental.plugins.sablier.version=v1.4.0-beta.6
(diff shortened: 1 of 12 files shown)


valexz:
It is my first experience with public PR, and seems i  did something wrong, when tried to make diff clearer for `beta `branch`. Should I return all to the starting point, or i may just create a new PR based on beta branch?

PR #179, opened 2023-09-13
fix(providers/kubernetes): Added QPS and Burst tweaks for client-side throttling

valexz:
Hello

The reasoning behind this PR is to enable the adjustment of QPS and Burst configuration parameters in the client-go library.

This will help avoid constant throttling issues when dealing with over 100 K8S deployments. The default values are insufficient for such a large number of deployments.

How client-side throttling works:
// NewTokenBucketRateLimiter creates a rate limiter which implements a token bucket approach.
// The rate limiter allows bursts of up to 'burst' to exceed the QPS, while still maintaining a
// smoothed qps rate of 'qps'.
// The bucket is initially filled with 'burst' tokens, and refills at a rate of 'qps'.
// The maximum number of tokens in the bucket is capped at 'burst'.

It is a new PR with clearer diff with beta branch, instead of broken #178 
I hope that working with that Pull Request will be more convenient

### DIFF
--- a/app/providers/kubernetes.go
+++ b/app/providers/kubernetes.go
@@ -12,6 +12,8 @@ import (
 	core_v1 "k8s.io/api/core/v1"
 
 	"github.com/acouvreur/sablier/app/instance"
+	providerConfig "github.com/acouvreur/sablier/config"
+	log "github.com/sirupsen/logrus"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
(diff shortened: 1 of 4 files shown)


acouvreur:
Thanks a lot @valexz 
Really appreciate it

acouvreur:
Because you changed the configuration, you will need to update the tests accordingly

Take a look at the config tests please, so we make sure we can load the new configuration through config file, cli arguments and env var

PR #227, opened 2023-11-29
feat(kubernetes-provider): get groups from statefulsets

massimeddu-sj:
This PR adds the support for statefulsets when getting the groups on Kubernetes.


acouvreur:
Thank you for your contribution :)

PR #241, opened 2023-12-26
fix kubernetes with traefik

tomaszduda23:
Address https://github.com/acouvreur/sablier/issues/142, https://github.com/acouvreur/sablier/issues/143, https://github.com/acouvreur/sablier/issues/131 and https://github.com/acouvreur/sablier/issues/62

### DIFF
--- a/plugins/traefik/main_test.go
+++ b/plugins/traefik/main_test.go
@@ -66,6 +66,25 @@ func TestSablierMiddleware_ServeHTTP(t *testing.T) {
 			},
 			expected: "response from sablier",
 		},
+		{
+			name: "sablier service is ready but 503",
+			sablier: sablier{
+				headers: map[string]string{
+					"X-Sablier-Session-Status": "ready",
+				},
+				body: "response from sablier",
+			},
+			fields: fields{
+				Next: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					w.WriteHeader(http.StatusServiceUnavailable)
+				}),
+				Config: &Config{
+					SessionDuration: "1m",
+					Dynamic:         &DynamicConfiguration{},
+				},
+			},
+			expected: "response from sablier",
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


acouvreur:
Hi @tomaszduda23 thank you for your contribution, I'll look into the details soon!

acouvreur:
After a quick reading, here's what I understand of your changes:

1. Sablier states that the backend is ready
2. You issue a request yourself to see if the underlying backend is in fact ready (sync between observed state and actual state)
3. If it's not ready
    1. If it's in the case of a blocking request you issue a redirect
    2. If it's in the case of a dynamic request you serve the response from sablie

Makes sense to me!
However I'll need your input and description on what you've done so we're sure everything is correct.

Also, check the builds failure, some tests appears to be broken. You might want to add test to cover all use cases you added.

Thanks for your contribution!

acouvreur:
And it seems like the integration test for e2e worked!
image

Good job!

tomaszduda23:
> * Sablier states that the backend is ready
> 
> * You issue a request yourself to see if the underlying backend is in fact ready (sync between observed state and actual state)
> 
> * If it's not ready
>   
>   1. If it's in the case of a blocking request you issue a redirect
>   2. If it's in the case of a dynamic request you serve the response from sablie

The request is issued to check if 503 was generated inside traefik due to missing server or by server itself. Traefik is not able to attach server to existing router. When scaling from 0 to 1 client has to make one extra request. Otherwise there will be 503.

tomaszduda23:
Tests are fixed. The rest of fails in pipeline seems to be related to permissions.

plugins/traefik/main_test.go
```diff
@@ -61,6 +67,90 @@ func TestSablierMiddleware_ServeHTTP(t *testing.T) {
...
+					httptrace.ContextClientTrace(r.Context()).WroteHeaders()
+					fmt.Fprint(w, "response from service")
+				}),
+				Config: &Config{
+					SessionDuration: "1m",
+					Blocking:        &BlockingConfiguration{},
+				},
+			},
+			expected: "response from sablier",
+			// is this correct for blocking? I would expect to get error
```
tomaszduda23, line 132: This seems to be wired to return 200 and json but this is how it was working before as well.

acouvreur:
I would like to merge this into beta, can you try and rebase your branch into beta first ?

tomaszduda23:
done

PR #242, opened 2023-12-28
Websocket PoC

tomaszduda23:
It could be used to implement websocket https://github.com/acouvreur/sablier/issues/21

### DIFF
--- a/plugins/traefik/main.go
+++ b/plugins/traefik/main.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptrace"
+	"strings"
 )
 
 type SablierMiddleware struct {
@@ -44,6 +45,11 @@ func (sm *SablierMiddleware) ServeHTTP(rw http.ResponseWriter, req *http.Request
 	}
 
 	defer resp.Body.Close()
+
+	if isWebsocketRequest(req) {
+		// TODO dynamic make no sense for websocket.
+		fmt.Println("=== websocket request")
+	}
 
 	conditonalResponseWriter := newResponseWriter(rw)
 
@@ -114,6 +120,7 @@ func (r *responseWriter) Write(buf []byte) (int, error) {
 }
 
 func (r *responseWriter) WriteHeader(code int) {
+	fmt.Println("=== code", code)
 	if r.ready == false && code == http.StatusServiceUnavailable {
 		// We get a 503 HTTP Status Code when there is no backend server in the pool
 		// to which the request could be sent.  Also, note that r.ready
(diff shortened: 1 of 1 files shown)


acouvreur:
I would like to merge this feature into beta first, can you try to rebase your branch into beta first please ?

tomaszduda23:
This is just proof of concept to show how it could be done. It was never completed. I could make it work but https://github.com/acouvreur/sablier/pull/241 needs to be merged first.

acouvreur:
I've squashed your commits from the previous PR, sorry if that causes issues on your side :D

acouvreur:
> This is just proof of concept to show how it could be done. It was never completed. I could make it work but #241 needs to be merged first.

Maybe we can mark the PR as draft?

DrummyFloyd:
do this PR was merge into Beta branch ?

tomaszduda23:
I've not completed it yet.

DrummyFloyd:
@tomaszduda23 do you need help =) i'll be glad to help

tomaszduda23:
The help would be great. You can open new PR based on that. I can help to review.
This PR just checked if this can be implemented. The logic which check if connection is still active is missing. The idea was:
1. Create goroutine during websocket request
2. the goroutine would get notification from Read/Write/Close and make sure that backend is scaled up
The goroutine is needed since http request is much heavier than tcp request(websocket).

DrummyFloyd:
what is your test case scenario ?

DrummyFloyd:
@tomaszduda23 @acouvreur 
when the `HiJack` function occur ? 
if i get it 

1.  you have a traefik route which point to an ws endpoint => whoami/echo 
2. sablier should scale up the pod/container to 1 if it's down => Seems OK to this part but return `200` instead of `101`
3. go routine should check if there is any activity on the WS endpoint
  3.1 if activity => no scale down (eg: `wscat -c localhost:8080/echo`)
  3.2 if no activity scaledown 
  ATM, it's scaledown no matterwhat, because only IngressRoute is available for middleWare nor IngressRouteTCP https://github.com/traefik/traefik/issues/10593 , and this trick will make work `websocket trhough sablier`
  
  am i right ? 
  will try to do this week end

tomaszduda23:
http is based on TCP. After http negotiation "hijack" happens and exactly the same connection is used for websocket. You do not need tcp middleware at all.
The connection part is already implemented. You just need to deal with those:
```
fmt.Println("=== websocket close") <- scale down
fmt.Println("=== websocket write", len(b)) <- keep up
fmt.Println("=== websocket read", len) <- keep up
```
Scale up is already implemented too.

Make the tests first. Than the rest will be clear when you see the logs.

tomaszduda23:
> * sablier should scale up the pod/container to 1 if it's down => Seems OK to this part but return `200` instead of `101`

in case of websocket only blocking strategy make sense. If user select dynamic strategy and service is scaled down I would return 503 probably.

DrummyFloyd:
@tomaszduda23 
thank you for you help 

iit seems to work , will make a bit a of cleaning , before opening a new PR =)

acouvreur:
Should be re-opened in https://github.com/sablierapp/sablier-traefik-plugin

PR #275, opened 2024-04-13
feat(traefik) allow websocket traffic

DrummyFloyd:
as @tomaszduda23  mentionned, i reoppend a PR
to make some reviews

atm

* docker test passed
* swarm (need to fix on my own swarm )
* Kube => bad request (issue on kube manifest i think

but tested manually

with docker 

i can reach `ws://localhost:8080/echo `
container is ready , i can talk with the `whoami WS` even, it's not shtudown , once i stop , after `session-duration` hte container i stopped


tomaszduda23:
> * Kube => bad request (seems to be a race conditions, because runing test mannualy seems to work `302 error on first try`) 

This is by design you need to follow redirection. Explained in https://github.com/acouvreur/sablier/pull/241

plugins/traefik/main.go
```diff
@@ -136,11 +164,132 @@ func (r *responseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
...
+	return err
+}
+
+func monitorWebSocketActivity(sablierRequest *http.Request, sm *SablierMiddleware) {
+	duration, err := time.ParseDuration(sm.config.SessionDuration)
+	if err != nil {
+		fmt.Fprintln(os.Stdout, []any{`Error parsing sessionDuration: %v`, err}...)
+		return
+	}
+	alertTime := duration - (duration * 5 / 100) // Calculate alert time at 95% of the total duration
```
tomaszduda23, line 259: I would do just half of the value which is simpler and more reliable imho.

plugins/traefik/main.go
```diff
@@ -136,11 +164,132 @@ func (r *responseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
...
+	for {
+		select {
+		case event := <-wsEventChan:
+			switch event {
+			case WebSocketRead, WebSocketWrite:
+				activeDuringAlert = true // Mark that there was activity during the alert period
+
+			case WebSocketClose:
+				fmt.Println("WebSocket closed")
+				activeDuringAlert = false // Do not reset ticker on close
```
tomaszduda23, line 275: you can complete go routine here. No need to wait for ticker.

plugins/traefik/main.go
```diff
@@ -8,19 +8,32 @@ import (
...
 
+type WebSocketEvent int
+
+const (
+	WebSocketRead WebSocketEvent = iota
+	WebSocketWrite
+	WebSocketClose
+)
+
+var wsEventChan = make(chan WebSocketEvent, 10)
```
tomaszduda23, line 24: why 10? Why not 5, 16 or 11 :grin:

plugins/traefik/main.go
```diff
@@ -47,6 +62,13 @@ func (sm *SablierMiddleware) ServeHTTP(rw http.ResponseWriter, req *http.Request
 
 	conditonalResponseWriter := newResponseWriter(rw)
 
+	if isWebsocketRequest(req) {
+		// FIXME dynamic make no sense for websocket since client return error
+		fmt.Println("=== websocket request")
+		go monitorWebSocketActivity(sablierRequest, sm)
```
tomaszduda23, line 68: I was thinking about this approach also. You create go routine per connection. I wondered if this would scale up. I didn't check details though.

plugins/traefik/main.go
DrummyFloyd: tbh i thought i removed this value ^^ should be `nil`

plugins/traefik/main.go
DrummyFloyd: i do this, 3-5 % of the duration , because, i need to trig Sablier but not on every  `event`, with this mechanism, i can refresh the scaling up( if i don't the Sablier scale down) with http

and i do not close on `WebSQocketClose` directly because u still have to wait the session duration

plugins/traefik/main.go
DrummyFloyd: check my comment above

plugins/traefik/main.go
DrummyFloyd: but the half duration make sense to me too

plugins/traefik/main.go
DrummyFloyd: it should scalre correctly , maybe in next iteration , we should add Mutex for security , but not really conformatable with it

DrummyFloyd:
> > * Kube => bad request (seems to be a race conditions, because runing test mannualy seems to work `302 error on first try`)
> 
> This is by design you need to follow redirection. Explained in #241

will check on that thank you

plugins/traefik/main.go
tomaszduda23: Now I get your point. It means that timeout will have random value depending on when the last request was done. It will be a little confusing. It would be better to make another request when connection is closed and cancel the timer. That way the timeout would be predictable from user point of view.

DrummyFloyd:
> > > * Kube => bad request (seems to be a race conditions, because runing test mannualy seems to work `302 error on first try`)
> > 
> > 
> > This is by design you need to follow redirection. Explained in #241
> 
> will check that! thank you
> 
> EDIT: tbh , i don't fully understand what happnd , can you please have a look @tomaszduda23

tomaszduda23:
It looks that you have to handle retry on client side https://github.com/trazfr/tcp-over-websocket/blob/fcb1b85e9fd0e282eee88815efeca8d729bd762b/httpClient.go#L86C14-L86C36

DrummyFloyd:
> It looks that you have to handle retry on client side https://github.com/trazfr/tcp-over-websocket/blob/fcb1b85e9fd0e282eee88815efeca8d729bd762b/httpClient.go#L86C14-L86C36

 Will have a look a it asap 

Thank you :)

tomaszduda23:
any progress on that?

DrummyFloyd:
Did not have time sorry to debug this 
Will try to give it a try next week

tomaszduda23:
> @tomaszduda23 can you please elaborate a bit what you means , by retry on client side ? because if i remembell well , this retry mechanism was already made your PR #241

It seems that go client won't handle redirect by itself.  You can reference this
https://github.com/trazfr/tcp-over-websocket/blob/fcb1b85e9fd0e282eee88815efeca8d729bd762b/httpClient.go#L86C14-L86C36. It works similar way to `curl` without `-L` option. Most clients will handle this by itself though. Redirect from http to https is pretty common.

DrummyFloyd:
> > @tomaszduda23 can you please elaborate a bit what you means , by retry on client side ? because if i remembell well , this retry mechanism was already made your PR #241
> 
> It seems that go client won't handle redirect by itself. You can reference this https://github.com/trazfr/tcp-over-websocket/blob/fcb1b85e9fd0e282eee88815efeca8d729bd762b/httpClient.go#L86C14-L86C36. It works similar way to `curl` without `-L` option. Most clients will handle this by itself though. Redirect from http to https is pretty common.

i tried some stuff, but everytime i brooke something else, dunno if i'll be able to finish that quickly

plugins/traefik/main.go
vbrandl: I wrote a proxy for Sablier to handle TCP only services and had a similar problem: How to prevent the service from shutting down, if there are only long running connections.

I don't check if there is actual activity on the connection, but I have a single task/process that will, as long as there is at least 1 active connection, ping the Sablier endpoint in regular intervals. Maybe something like this would work here, too:

https://github.com/vbrandl/sablier-proxy/blob/main/src/main.rs#L154

This way, you don't have to spawn another goroutine per connection, but there will be exactly 1 monitoring task/routine, which might scale better if there are many connections.

acouvreur:
Should be re-opened in https://github.com/sablierapp/sablier-traefik-plugin

PR #284, opened 2024-04-17
feat(plugin): add `proxywasm` plugin

acouvreur:
The `proxywasm` plugin is a WASM Filter following the ProxyWasm ABI Specification using the proxywasm go sdk

This allows extensibility with any reverse proxy who implements the ProxyWasm ABI Specification.

The current WASM Filter was successfully tested with APISIX and Envoy.


acouvreur (commented):
This is a first draft for the `proxywasm` plugin.

This plugin will allow a great extensibility and usage within the reverse proxy world.

More examples are to come with Nginx and Istio.

---

What's left to do:
- Add nginx and istio examples
- Add E2E tests
- Publish a WASM Docker Image that will be used with Istio (ci)
- Enhance documentation
- All examples should have a README on how to run them with a full execution example.

acouvreur:
There's one remaining issue before marking this PR as ready:

It's the way the `proxywasm.DispatchHttpCall` function works depending on the reverse proxy.

The signature of the function is the following:

```go
// DispatchHttpCall is for dispatching HTTP calls to a remote cluster. This can be used by all contexts
// including Tcp and Root contexts. "cluster" arg specifies the remote cluster the host will send
// the request against with "headers", "body", and "trailers" arguments. "callBack" function is called if the host successfully
// made the request and received the response from the remote cluster.
// When the callBack function is called, the "GetHttpCallResponseHeaders", "GetHttpCallResponseBody", "GetHttpCallResponseTrailers"
// calls are available for accessing the response information.
func DispatchHttpCall(
	cluster string,
	headers [][2]string,
	body []byte,
	trailers [][2]string,
	timeoutMillisecond uint32,
	callBack func(numHeaders, bodySize, numTrailers int),
) (calloutID uint32, err error) {
```

Istio expects the following arguments:
- The cluster in the following format: `"outbound|10000||sablier"`
- The following headers:
  - `":method": GET`
  - `":path": /api/strategies/...`
  - `":authority": sablier` 

Other reverse proxies (Ngin and APISIX) expect the following arguments:
- The cluster in the following format: `sablier:10000`
- The following headers:
  - `":method": GET`
  - `":path": /api/strategies/...`
  - `":authority": sablier:10000` 

Kong ref: https://github.com/Kong/ngx_wasm_module/blob/b19d405403ca6765c548e571010aea3af1accaea/src/http/proxy_wasm/ngx_http_proxy_wasm_dispatch.c#L255-L269
APISIX Ref: https://github.com/api7/wasm-nginx-module/blob/0b4b31d6ecfdbc587e8ea455ed6e920a98aadff1/src/http/ngx_http_wasm_call.c#L38-L44

---

So depending of the execution context, I have to provide different ways of build the configuration.
I could expose the cluster argument and authority arguments directly so I don't have to pre-bake them.

But I'm not really sure.

acouvreur:
Everything is almost ready:
- need to add e2e tests
- more documentation

PR #286, opened 2024-04-18
ci(docker): distribute build across multiple runners

acouvreur:
Docker build for multiplatform arch is too slow sequentially on the same runner.

The build is now distributed on multiple runners. Each architecure image is pushed as a digest and is reconciled into a single image afterwards.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,7 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version: ^1.21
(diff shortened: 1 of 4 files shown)


acouvreur:
Conclusion on how it went so far:
- Docker build time is reduced from an average of 25 minutes to 14 minutes
- The amd64 docker build takes 2 minutes maximum
- arm builds take a lot of time due to emulation being very slow through qemu
- You can check availability for arm runners here: https://github.com/github/roadmap/issues/836

PR #290, opened 2024-04-24
Fix syntax error in Caddy dockerfile.

valankar:


acouvreur:
Thanks for the PR @valankar !

Issue #287, opened 2024-04-19
Support for docker-socket-proxy


thielj:
Hi,

can you add support for controlling docker through a network proxy, instead of requiring direct access to the socket, please? Similar to what setting `DOCKER_HOST` or using docker -H `<endpoint>` would do.

For example, through a configuration setting like this:

```
    endpoint: "unix:///var/run/docker.sock"     # Default, listen to the UNIX Docker socket
    #endpoint: "tcp://docker-socket-proxy:2375"   # Use a network proxy instead
```

Ideally, this would be configurable per service or middleware, which would allow managing a containers on different hosts.

See https://github.com/Tecnativa/docker-socket-proxy and https://flast101.github.io/docker-privesc/ for some background info.

acouvreur:
Hello!

Yes the library is configured to support all environnement variables.
Yes we could add it to the docs!

Sblop:
Hi. While this works, the sablier container now runs at 100% CPU. Logs doesnt assist.

acouvreur:
Closing this in favor of #272

Issue #272, opened 2024-04-04
high cpu usage with only two services managed


cyrinux:
**Describe the bug**
High CPU usage 

**Context**
 - Sablier version:  1.6.0
 - Provider: docker v20.10.23
 - Reverse proxy: traefik v2.11
 - Sablier running inside a container

**Expected behavior**
Low cpu usage

**Additional context**
I manage 2 services in a big docker-compose files with sablier, and when I do a top I can see sablier process very hard with my tiny cpu.

Is it a know issue ? I can see there is a full rewrite PR in progress, this can help ? (I didnt dig in the code actually)

kraoc:
I'll add a comment here.

I have several VMs with Sablier.
I notice an increase in its cpu consumption without apparent reason.

When I restart Sablier, its cpu consumption returns to normal.

acouvreur:
Hello @kraoc, do you have more inputs on that ?

If you have maybe CPU usage graphs, or specific events after which you noticed that ?

I will investigate this issue for sure when I have time, but more input to narrow down the issue will help!

kraoc:
I'll try to add more usefull debug tips :p
(but I need to dig a bit)

My first think point to a potential leak on first start (eg. on a vm boot) ; because after a restart of sablier it's cpu consumption is ok.
Dunno if I'm making myself clear :p

luked34:
I have the same issue. I am running in Docker Desktop on Ubuntu, with the CPU limit under "Resources" set to 4.

After restarting the container the CPU usage is almost zero. But about 4 minutes later it ramps up to just over 100% (the maximum available is 400%). This happens whether or not I send any requests through Sablier. The usage is coming from the main Sablier container, not the proxy (I use Caddy).

It seems to work fine even when the CPU usage is high. I can't see anything useful in the logs, but let me know if there's anything else I can provide.

cyrinux:
I still have the bug too, even with the last version, always same behavior, after restart its cool. And when I hit one of the endpoint manage by sablier then the process consumne lot of cpu.
I tried to tweak some sablier params but this dont really help.

acouvreur:
I'd be curious to have more information to try to reproduce the issue.
What I'd need would be the following:
- OS (Windows, Mac, linux)
- OS Version
- Where is Sablier running ? (binary, docker, and if it's docker, is it in docker desktop ?)
- The configuration file for sablier
- The endpoint you are reaching and the parameter used after which you see this CPU usage.

[...]

PR #315, opened 2024-06-04
Update docker_classic.go with delay

Sblop:
This change addresses the issue reported in Issue #287, where the CPU load reaches 100% when the DOCKER_HOST environment variable is set to a value other than the default /var/run/docker.sock. The infinite loop in the NotifyInstanceStopped method was causing excessive CPU usage due to continuous polling without any delay. Introducing a small delay (100 milliseconds) in the loop reduces CPU load while maintaining functionality.  This modification ensures that the event loop does not consume excessive CPU resources, especially when connected to a remote Docker host.

### DIFF
--- a/app/providers/docker_classic.go
+++ b/app/providers/docker_classic.go
@@ -6,7 +6,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
-
+        "time"
 	"github.com/docker/docker/api/types/container"
 
 	"github.com/acouvreur/sablier/app/instance"
--- a/app/providers/docker_classic.go
+++ b/app/providers/docker_classic.go
@@ -152,8 +152,14 @@ func (provider *DockerClassicProvider) NotifyInstanceStopped(ctx context.Context
 				log.Debug("provider event stream closed")
 				return
 			}
+            		if err != nil {
+                		log.Error("error while processing Docker events:", err)
+            		}			
 		case <-ctx.Done():
 			return
+        	default:
+            		// Add a small delay to prevent high CPU usage
+            		time.Sleep(100 * time.Millisecond)
 		}
 	}
 }


acouvreur:
Hello @Sblop !

Thank you for your contribution!

---
Did you successfully run this version of the code and saw significant CPU usage reduction ?
I can produce a docker image for you to try if needed.

---
Can you please give more details on why do you think this is the root cause of the CPU usage ?


Because I'd believe a single call is done, and then it is waiting for events, right ?

So how come the select statement is looping out of control ?

Sblop:
In my initial attempt, this seemed to bring down the CPU, but I believe I was mistaken... After first container has been started, the CPU stays high, using the DOCKER_HOST env. The debug log stays the same:

```
time="2024-06-06T12:13:16+02:00" level=debug msg="map[]"
time="2024-06-06T12:13:18+02:00" level=debug msg="map[]"
time="2024-06-06T12:13:20+02:00" level=debug msg="map[]"
time="2024-06-06T12:13:22+02:00" level=debug msg="map[]"
time="2024-06-06T12:13:24+02:00" level=debug msg="map[]"
time="2024-06-06T12:13:26+02:00" level=debug msg="map[]"
time="2024-06-06T12:13:28+02:00" level=debug msg="map[]"
time="2024-06-06T12:13:30+02:00" level=debug msg="map[]"
time="2024-06-06T12:13:32+02:00" level=debug msg="map[]"
time="2024-06-06T12:13:34+02:00" level=debug msg="map[]"
```

Do you have any suggestions what might cause this? 
I belive it will be fine if you archive this pull request.

acouvreur:
I believe that these logs are from the group discovery. Which queries all your containers every 2 seconds.

Check this:

https://github.com/acouvreur/sablier/blob/main/app/sessions/sessions_manager.go#L64

And

https://github.com/acouvreur/sablier/blob/main/app/sessions/groups_watcher.go

acouvreur:
If you can enable some monitoring and send some report of the CPU usage, it could be very interesting.

I can try to have that kind of performance check in the future. That might interesting to see the impact of let's say, an upgrade on golang, or a docker package, whatever.

acouvreur:
Hello @Sblop I will close this issue in favor of #329 

You can also take a look at #272 for the whole investigation.

PR #329, opened 2024-06-27
fix(docker): ensure connection is established with host

acouvreur:
The `docker` and `docker_swarm` providers have been patched to ensure that the connection is properly established upon starting.

If the docker host is not available at starting time, then the application will stop. This will prevent from trying to register to the event stream on a non working client.


acouvreur:
Fixes #272

PR #338, opened 2024-06-29
Update traefik.md

tomaszduda23:
blocking for traefik is already supported


acouvreur:
Thanks @tomaszduda23 !

Issue #388, opened 2024-09-06
Cloudflare and cached loading page resulting in javascript/mime errors


valankar:
**Describe the bug**
I'm running Sablier via Caddy behind a Cloudflare tunnel. My app is a FastAPI/NiceGUI Python app, which retrieves .js files. It generally works well most of the time, but sometimes I get a bunch of errors with the .js files actually returning the Sablier loading page HTML. It looks like the following:

![js errors](https://github.com/user-attachments/assets/ccd73c7b-2b5f-439b-ba21-8e336b1de98e)

When I go to inspect those .js files in the browser, it appears the text is actually the Sablier loading page:

![js errors 2](https://github.com/user-attachments/assets/60a82057-60dd-4870-8049-898968b99e9f)

```
    <h1><span>Starting </span> <span class="error_code"></span>...</h1>
    <p class="output"><span>Your instance(s) will stop after 6 hours of inactivity</span>.</p>
...
```

What I think is happening is Cloudflare is caching that loading page and returning it for all requests, even after the container is up. I notice the page doesn't have any headers to affect caching. Shouldn't it have some to instruct clients not cache it?

**Context**
 - Sablier version: v1.8.0-beta.18
 - Provider: Docker version 27.2.0, build 3ab4256
 - Reverse proxy: caddy
 - Sablier running inside a container? yes

**Expected behavior**
Maybe the Sablier loading page should specifically have no cache headers.

acouvreur:
Hello @valankar that's actually a pretty good point!

Do you think you could submit such fix ?

valankar:
Sure I'll do some tests with https://github.com/acouvreur/sablier/blob/main/docs/themes.md first.

acouvreur:
@valankar any feedback on the released version ? Does it work as expected ?

valankar:
It still unfortunately happens around 5% of the time, usually when I hit my website on my phone (Safari). It's rather difficult to debug. I know that NiceGUI/FastAPI use websockets, so perhaps using dynamic strategy is not the right thing. Everything works fine with blocking strategy.

PR #389, opened 2024-09-06
Provide Cache-Control header on dynamic page to prevent caching by services like Cloudflare.

valankar:


acouvreur (approved):

acouvreur:
Awesome work @valankar I can merge it right now and it will trigger a new beta release that you can use.

acouvreur:
Should fix #388

PR #274, opened 2024-04-10
feat(charts): sablier helm chart for kube deployement

DrummyFloyd:
Hi , like i said i n the issue, seems to be a good idea to have a chart to easily deploy sablier

* templates
* unittest
* ci-lint test (ct lint/install  and helm unittest)
* readme generation docs with helm-docs
* ct-releaser => may not fully functionnal

NB: not very used to work with gh-actions, double check ^^

Closed: #271

PR #403, opened 2024-09-16
271 kubernetes helm chart

acouvreur:
Supersed #274


helm/templates/rbac.yaml
```diff
@@ -0,0 +1,32 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: sablier-{{ .Release.Name }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: sablier-{{ .Release.Name }}
+subjects:
+  - kind: ServiceAccount
+    name: sablier
```
MrLuje, line 31: The `sa` name doesn't match the one created in the other file

```suggestion
    name: sablier-{{ .Release.Name }}
```

acouvreur:
Hello everyone, as mentioned I will be moving this to a new organization. So I've actually baked the helm chart: https://github.com/sablierapp/helm-charts

You can start testing it! I'm sure there will be some issues but it's a first :)

PR #452, opened 2024-11-17
fix(kubernetes): support public labels

### LINKED ISSUE #449: (Kubernetes) labels not public
**Describe the bug**

Sablier uses private [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) (private because there is no `/` in the key). Example:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: whoami
  labels:
    app: whoami
    sablier.enable: "true"
    sablier.group: mygroup
…
```

**Context**
 - Sablier version: 1.8.1
 - Provider: any Kubernetes
 - Reverse proxy: _not applicable_
 - Sablier running inside a container? _not relevant_

**Expected behavior**
The label keys for Sablier are public; for example:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: whoami
  labels:
    app.kubernetes.io/name: whoami # also good practice, but not core to THIS issue
    sablierapp.dev/enable: "true"
    sablierapp.dev/group: mygroup
…
```

Using public label keys is good practice, helps people find the documentation a little more easily, and avoids conflicting with organizations' own policies and conventions for private labels.

**Additional context**
Typically when Kubernetes-compatible projects adopt the official label convention(s), they add support for the new way; then they deprecate the old way (using private label keys), but continue to support it for many releases. The documentation recommends the new-style, public label keys.

Read https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set for Kubernetes' documentation about public vs. private label keys.

acouvreur:
closes #449

### DIFF
--- a/app/providers/kubernetes/list.go
+++ b/app/providers/kubernetes/list.go
@@ -3,23 +3,31 @@ package kubernetes
 import (
 	"context"
 	"github.com/sablierapp/sablier/app/discovery"
-	"github.com/sablierapp/sablier/app/providers"
 	"github.com/sablierapp/sablier/app/types"
 	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/api/apps/v1"
 	core_v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"strconv"
-	"strings"
 )
 
-func (provider *KubernetesProvider) InstanceList(ctx context.Context, options providers.InstanceListOptions) ([]types.Instance, error) {
-	deployments, err := provider.deploymentList(ctx, options)
+const (
+	LabelEnable                      = "sablierapp.dev/enable"
+	LabelGroup                       = "sablierapp.dev/group"
+	LabelGroupDefaultValue           = "default"
+	LabelReplicas                    = "sablierapp.dev/replicas"
+	LabelReplicasDefaultValue uint64 = 1
+)
+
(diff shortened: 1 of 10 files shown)


sftim (commented):
LGTM

acouvreur:
> LGTM

There's an issue with a nil pointer, I have to fix that

app/providers/kubernetes/list.go
```diff
@@ -3,33 +3,51 @@ package kubernetes
...
+	requirement, err := labels.NewRequirement(LabelEnable, selection.Equals, []string{"true"})
+	if err != nil {
+		return nil, err
+	}
+	requirementDeprecated, err := labels.NewRequirement(discovery.LabelEnable, selection.Equals, []string{"true"})
+	if err != nil {
+		return nil, err
+	}
+	selector := labels.NewSelector()
+	selector = selector.Add(*requirement, *requirementDeprecated)
```
BapRx, line 48: If you merge the 2 selector type here, the deployments will need to have both legacy and new labels because the selector would result in `sablier.enable=true,sablierapp.dev/enable=true`.

I don't know how to handle this, either deprecate old selectors or add a flag to toggle legacy/new (or make it customizable?)

BapRx: After a bit more tweaking I got it to work, I added a bunch of Trace logging to help me find the root cause and it was mainly due to old `discovery.LabelGroup` usage.

I rebased main onto my branch (449-xxx) so I cannot open a PR on yours but you can see my diff here: https://github.com/sablierapp/sablier/compare/main...BapRx:sablier:449-kubernetes-labels-not-public

acouvreur: Awesome, I will check it out.

acouvreur:
Superseded by #1114

PR #461, opened 2024-11-28
feat: add refreshFrequency option for blocking strategy

nickalie:
Currently, for initial start blocking strategy always takes at least 5 seconds, before bypassing the request to the target container. Though the container itself may start in half a second, blocking strategy still waits for 5 seconds due to hard-coded refresh frequency.

The PR adds customizable `refreshFrequency` option to the blocking requests. Works in similar way as `refreshFrequency` option for the dynamic strategy. For instance, if you set `refreshFrequency` to 100ms and your container starts in 500ms, the end user will receive response for the very first request in about 600-700ms instead of 5000ms for blocking strategy


acouvreur:
Hey @nickalie thanks for that pull request!

acouvreur:
Some issues 
image

Probably from parsing or default value ?

acouvreur:
@nickalie any update on this pull request?

Is it ok if I take it over ?

nickalie:
> @nickalie any update on this pull request?
> 
> Is it ok if I take it over ?

Yes, please. About 3 months ago I fixed all tests and now looks like there were multiple critical updates in the upstream, I'm not sure when I'll be able to solve the conflicts

acouvreur:
> > @nickalie any update on this pull request?
> > Is it ok if I take it over ?
> 
> Yes, please. About 3 months ago I fixed all tests and now looks like there were multiple critical updates in the upstream, I'm not sure when I'll be able to solve the conflicts

Will do, indeed the whole project went under very active development. So I'll do it no problem.

PR #486, opened 2025-01-14
fix(kubernetes): early-exit on error

BapRx:
Helped me troubleshoot a SIGSEGV error:

- before: `panic: runtime error: invalid memory address or nil pointer dereference`
- after: `panic: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory`


acouvreur:
Nice catch!

PR #536, opened 2025-02-25
Feature/docker reconnect

### LINKED ISSUE #535: Improve Docker socket connection resiliency when using socket proxy
## Issue Description

When using Sablier with a Docker socket proxy (rather than direct socket access), the application fails to properly handle connection interruptions. Specifically, if the socket proxy terminates a connection due to timeouts or other reasons, Sablier logs an error but continues to report itself as healthy while being unable to execute Docker operations.

### Current Behavior

1. Sablier establishes a connection to Docker via a socket proxy
2. If the socket proxy times out or terminates the connection, Sablier logs:
   ```
   ERR docker/docker.go:155 event stream error provider=docker error="unexpected EOF"
   ERR docker/docker.go:148 event stream closed provider=docker
   ```
3. Sablier's health check (`/health` endpoint) continues to return a 200 status code with "OK"
4. Subsequent Docker operations fail silently, and users must manually restart the Sablier container

### Expected Behavior

1. Sablier should attempt to reconnect to the Docker daemon when the connection is lost
2. The health check should verify Docker connectivity, not just API responsiveness
3. If reconnection fails after several attempts, Sablier should either:
   - Update its health status to unhealthy
   - Log a clear error that the Docker connection is permanently lost
   - Automatically restart itself (if possible)

## Reproduction Steps

1. Configure Sablier to use a socket proxy with a timeout value (e.g., `PROXY_READ_TIMEOUT=8000`)
2. Wait for the timeout to occur
3. Observe Sablier logs showing "event stream error" and "event stream closed"
4. Verify that `/health` still returns 200 OK
5. Attempt to start a container through Sablier, which will fail

## Technical Details

This issue stems from how Sablier handles the Docker client connection in the provider implementation. In the Docker provider, there appear to be two key issues:

1. **Event stream resilience**: The `NotifyInstanceStopped` method establishes a connection to the Docker events API, but doesn't automatically reconnect if this connection is lost.
2. **Docker client reuse**: The Docker client is created once during provider initialization and reused, but there's no mechanism to verify its connectivity or recreate it if it becomes invalid.

## Proposed Solutions

1. **Implement automatic reconnection to Docker daemon**:
   - Add a reconnection loop in the event stream handler
   - Periodically verify Docker connectivity and recreate the client if needed

2. **Enhance health check**:
   - Update the `/health` endpoint to verify Docker connectivity
   - Add a new `/health/docker` endpoint specifically for Docker connectivity

3. **Connection tracking**:
   - Track the most recent successful Docker operation
   - If operations fail or too much time passes since the last successful operation, attempt to recreate the connection

## Environment Information

- Sablier version: 1.8.4
- Docker version: 27.5.1
- Docker API version: 1.47
- Socket proxy: Used with `PROXY_READ_TIMEOUT=8000`

## Additional Context

This issue becomes particularly problematic in production environments where Sablier is a critical component for managing containers, as it requires manual intervention to recover from connection issues.

mg-dev25:
## Fixes #535

This PR implements the improvements for Docker connection resiliency as described in issue #535, particularly when using a Docker socket proxy.

### Changes:

- Added reconnection logic with exponential backoff for Docker event streams
- Fixed "unexpected EOF" handling with proper reconnection attempts
- Implemented Docker connectivity verification in health checks
- Added detailed logging for connection interruptions and successful reconnections
- Fixed linting issues in multiple files

### Solution Implementation:

1. **Event Stream Resilience**:
   - Added a reconnection loop with exponential backoff in the event stream handler
   - Implemented proper error handling for "unexpected EOF" and other connection issues
   - Added clear logging of reconnection attempts and successes

2. **Health Check Enhancement**:
   - Updated the `/health` endpoint to verify Docker connectivity
   - Added Docker API version to health check responses
   - Health check now returns unhealthy status when Docker connection is lost

3. **Code Quality Improvements**:
   - Fixed type naming to avoid stutter (DockerClassicProvider → ClassicProvider)
   - Improved control flow in health check logic
   - Fixed various linting issues throughout the codebase

### Testing Done:

- Tested reconnection by manually interrupting Docker socket proxy
- Verified proper reconnection with Docker daemon after connection loss
- Validated health check properly reports Docker connectivity issues
- Confirmed linting issues are resolved


app/sablier.go
```diff
@@ -37,6 +37,16 @@ func Start(ctx context.Context, conf config.Config) error {
 		return err
 	}
 
+	// Extract Docker ping function for health checks if we're using Docker
```
acouvreur, line 40: We should instead add a new method to every provider to check the current provider health.

I have the following names in mind:
- `Probe`
- `Ping`
- `Health`

Then this function should only call the provider healthchecking method.

app/sablier.go
```diff
@@ -112,7 +122,7 @@ func Start(ctx context.Context, conf config.Config) error {
 		SessionsConfig:  conf.Sessions,
 	}
 
-	go server.Start(ctx, logger, conf.Server, strategy)
+	go server.Start(ctx, logger, conf.Server, strategy, dockerPingFunc)
```
acouvreur, line 125: The server should not have a provider specific function, instead it should be a provider wide health check function.

Related to my previous comment.

config/provider.go
```diff
@@ -33,6 +44,11 @@ func NewProviderConfig() Provider {
 			Burst:     10,
 			Delimiter: "_", //Delimiter used for namespace/resource type/name resolution. Defaults to "_" for backward compatibility. But you should use "/" or ".".
 		},
+		Docker: Docker{
+			ReconnectInitialDelay: 1 * time.Second,
+			ReconnectMaxDelay:     30 * time.Second,
+			ReconnectMaxAttempts:  0, // 0 means unlimited
```
acouvreur, line 50: Maybe -1 should mean unlimited ?
0 would mean no retry at all

internal/api/health.go
```diff
@@ -3,17 +3,26 @@ package api
...
+	"sync"
+	"time"
 
 	"github.com/gin-gonic/gin"
 )
 
 type Health struct {
 	TerminatingStatusCode int `description:"Terminating status code" json:"terminatingStatusCode,omitempty" yaml:"terminatingStatusCode,omitempty" export:"true"`
 	terminating           bool
+	dockerPingFunc        func(context.Context) error
```
acouvreur, line 15: I would change to a non-specific provider func.

Everything related to docker specifically should be drop

internal/api/health.go
```diff
@@ -23,18 +32,90 @@ func (h *Health) WithContext(ctx context.Context) {
...
+
+	// Check Docker health status if we're using Docker provider
+	h.dockerPingMutex.RLock()
+	dockerHealthy := h.dockerPingStatus
+	dockerMessage := h.dockerPingMessage
+	lastCheck := h.dockerPingLastCheck
+	h.dockerPingMutex.RUnlock()
+
+	// If Docker is not healthy and Docker provider is being used, return unhealthy
+	if h.dockerPingFunc != nil && !dockerHealthy {
```
acouvreur, line 93: Call the `providerHealthFunc` and if error is not nil, write the error message on the body.

I would remove the headers you added completely

internal/server/server.go
```diff
@@ -32,12 +32,12 @@ func Start(ctx context.Context, logger *slog.Logger, serverConf config.Server, s
...
+	r := setupRouter(ctx, logger, serverConf, s, dockerPingFunc)
 
-	var server *http.Server
-	server = &http.Server{
-		Addr:    fmt.Sprintf(":%d", serverConf.Port),
-		Handler: r,
+	server := &http.Server{
+		Addr:              fmt.Sprintf(":%d", serverConf.Port),
+		Handler:           r,
+		ReadHeaderTimeout: 10 * time.Second, // Prevent Slowloris attacks
```
acouvreur, line 40: Could you make this change in another pull request, so we can properly track multiple changes

pkg/provider/docker/docker.go
```diff
@@ -127,34 +137,102 @@ func (p *DockerClassicProvider) GetState(ctx context.Context, name string) (inst
...
 func (p *DockerClassicProvider) NotifyInstanceStopped(ctx context.Context, instance chan<- string) {
-	msgs, errs := p.Client.Events(ctx, types.EventsOptions{
-		Filters: filters.NewArgs(
-			filters.Arg("scope", "local"),
-			filters.Arg("type", string(events.ContainerEventType)),
-			filters.Arg("event", "die"),
-		),
-	})
-	for {
+	reconnectDelay := p.reconnectInitialDelay
```
acouvreur, line 140: These changes are rather big, and before reviewing this bit, let's try to find some github project that uses the event endpoint and that does reconnect

acouvreur (changes requested):
Thanks @mg-dev25 for your contribution

Overall this changed is too much focused on docker for the healthcheck part where I think it could be implemented at the provider level.

Other providers can simply return true for the moment.

What do you think ?

mg-dev25:
It's great, I will work on it.

acouvreur:
@mg-dev25 let me know if you are working on it, otherwise I can also make a few adjustements.

But we should try not to. work together on the same things :)

mg-dev25:
Hi @acouvreur ,
I'm quite new to Git and wanted to clarify how you'd like me to proceed with the changes. Based on your feedback, I'm thinking of splitting the work into 3 separate PRs:
PR #1: Provider Health Interface

Add Health() method to the Provider interface
Implement provider-specific health checks for each provider type
Make the health check API provider-agnostic (removing Docker-specific code)
Update server and routes to use the generic provider health function

PR #2: Docker Reconnect Config Change

Change Docker reconnect attempts default from 0 to -1 for unlimited
Update condition check in reconnect logic to use >=0 instead of >0

PR #3: HTTP Security Headers

Add ReadHeaderTimeout: 10*time.Second to prevent Slowloris attacks
Keep this as a separate small PR as you suggested

Regarding your comment about the Docker reconnection logic: "These changes are rather big, and before reviewing this bit, let's try to find some github project that uses the event endpoint and that does reconnect"
I understand you'd like to see examples of proven implementation patterns before proceeding with the Docker event reconnection code. Would you prefer I research some reference implementations first? I could look for established projects that handle Docker event stream reconnection and propose a revised approach based on those patterns.
Would you like me to proceed with creating these separate PRs, or should we reconsider the reconnection implementation first?
Thanks for your guidance!

PR #580, opened 2025-04-06
installation.md: fix misspelled config filename

thomas-mc-work:
Took me a while to find the reason why my config didn't apply … :)


acouvreur:
Thanks!

Issue #148, opened 2023-04-07
Add `systemd` provider


tandy-1000:
**Describe the provider**
Systemd is used on NixOS to provide OCI containers, this is my usecase.
This would also open up Sablier to work for many other providers which already utilise systemd.

**Does the provider provides the following APIs**
*An instance is a generic term used to describe containers, pods, deployments, and more*

| API | Yes | No | I don't know |
| ---- | ---- | --- | -----------|
| Start an instance | ✅ | |
| Stop an instance | ✅ | |
| Get info about an instance | ✅ | |
| Listening for external events happening  |✅||

**Does the provider has a Go API SDK?**

https://github.com/coreos/go-systemd

**Does the provider can be started locally**
Yes.

acouvreur:
Thanks for your proposal @tandy-1000 !
I'm not really familiar with `Systemd`.

And you  say:
> Systemd is used on NixOS to provide OCI containers

Could you please provide an example of the scenario when you do it by hand ?

sftim:
https://mgdm.net/weblog/systemd-socket-activation/ explains a similar approach. I don't know which part of this might be useful to Sablier.

TheRedCyclops:
Support for systemd would be great, certain applications such as ollama provide systemd services for activation and unlike docker it is really easy writing your own units since they allow you to just execute an application directly on your system

acouvreur:
> I attempted to work on a systemd provider, to make Sablier work with Podman Quadlets (which use systemd under the hood). I drafted a PoC here: [#611](https://github.com/sablierapp/sablier/pull/611)
> 
> Theres two challenges with systemd:
> 
>     1. Unlike Docker, the dbus API doesn't really allow easy filtering on custom labels/tags. In fact the API won't return any custom attributes.
>        For example, in my implementation i add custom properties to my system unit, e.g.
> 
> 
> [Unit]
> After=...
> 
> [Service]
> ExecStart=...
> 
> [X-Sablier]
> Enable=true
> Group=test
> 
> Since `X-` prefixed options will be ignored and not returned by the dbus API, one would have to parse the units .ini file to access them. E.g. from the [docs](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html):
> 
> > Applications may use this to include additional information in the unit files. To access those options, applications need to parse the unit files on their own.

I'm fine with having to parse these files, do you see any issue with that?

[...]

PR #611, opened 2025-05-25
feat: add systemd provider

Tarow:
Addresses https://github.com/sablierapp/sablier/issues/148

As of now it's still WIP as tests will have to be added etc.


acouvreur:
Hey @Tarow !

Thanks for this nice contribution!

Do you think we could add tests using testcontainers ? That should work inside a clean ubuntu image right ? We could write some helper functions to manipulate the systemd files.

I will take a closer look to the files when I have time, I'm quite busy at the moment.

Tarow:
Getting systemd to work in Docker containers can be a bit tricky. AFAIK it only works in rootful containers and i'm not sure if/how good it will work with Testcontainers.
I can explore it a bit when i find time, but in case it isn't possible, writing a mock for the systemd interface might be the only option.

Tarow:
I've updated the provider to the latest interface contract, made a couple improvements and added tests.

Some background: 

Systemd will not report custom properties in unit files (e.g. `X-Sablier` section in this case) when using `systemctl show <unit>` and equivalent library functions.
So in order to get the custom sablier labels, the unit file itself has to be parsed.
The full flow is:
List units -> Get `FragmentPath` for each unit -> parse unit file -> extract sablier labels

To avoid excessive I/O and not parse hundrets of unit files every couple seconds, i added two additional mechanisms for the systemd provider:
1. A new option `provider.systemd.unit-patterns`: The provider will only look at unit files matching this pattern. For example, if only podman quadlets should be managed by sablier, it'd make sense to set this to `podman-*.service`.
Each inspected unit still needs appropriate sablier labels in a [X-Sablier] section (Enable, Group, ...)
2. Cache the sablier labels for each fragment path based on mtime/size. If those didn't change, the unit file is not parsed again. That only requires an `os.stat` and is a lot cheaper than reading and parsing the entire file every cycle.

Both of those together lower file I/O by a lot, so the provider should be a lot more efficient now.

acouvreur:
Hello @Tarow thank you for your contribution, can you add an example under `examples/systemd` ?

acouvreur:
Looks amazing @Tarow thanks a lot. I'm still reviewing your work. The example is making it really easy

acouvreur:
I realize that the tests are not being executed because we need to use SABLIER_SYSTEMD_INTEGRATION.

Can we find a way to always run those tests ? Or is the setup too complicated / impossible ?

acouvreur:
Maybe a dedicated github ci job:

```
  integration-systemd:
    name: Integration (systemd)
    runs-on: ubuntu-latest
    steps:
      # checkout + setup-go
      - run: sudo apt-get install -y dbus   # mock tests t.Skip without dbus-daemon
      - run: go test -tags=nomsgpack ./pkg/provider/systemd/...
        env:
          SABLIER_SYSTEMD_INTEGRATION: "1"
```

Tarow:
It is definitely possible to always run those. The setup wasn't super trivial (`init.sh` entrypoint for the testcontainer), but it should work consistently now.

The reason why i added the additional environment variable is, that all other integration tests for docker, podman, k3s have a fully isolated container that doesn't really touch the host.
Now for the systemd integration test the provider configures real cgroup resource limits (cpu quota, memory, I/O, ...) on the managed unit. Applying those settings requires write access to the cgroup hierarchy (can also be seen in the testcontainers instructions: https://hub.docker.com/r/jrei/systemd-ubuntu). Because the test writes to the hosts cgroupfs (and could technically change resource limits), i kept it opt-in. If you're fine with always running those anyways, the env variable and the extra GH action i created now could be removed.

.github/workflows/integration-systemd.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: "go.mod"
+      - name: Run systemd integration tests
+        env:
+          SABLIER_SYSTEMD_INTEGRATION: "1"
```
acouvreur, line 31: > Error: The actions actions/checkout@v7 and actions/setup-go@v5 are not allowed in sablierapp/sablier because all actions must be pinned to a full-length commit SHA.

you will need to use the pin with version in comment notation for those

Tarow: My bad, fixed that now

acouvreur (dismissed):
LGTM!

Waiting green worfklows, then merging!

acouvreur:
There might be some conflicts following https://github.com/sablierapp/sablier/pull/1087. Can you look into it @Tarow ? Sorry for the inconvenience

Tarow:
No worries, i'll check it out later.

acouvreur (dismissed):
LGTM

acouvreur (dismissed):

Issue #212, opened 2023-10-10
Feature Request: Ability to halt the container


Svenum:
I want just halt the container for faster wake up times. Maybe with a setting in sablier.yaml?
This should save CPU + its fast up and running.

acouvreur:
The underlying action for docker is the equivalent of `docker start` and `docker stop`.

By halting containers, do you mean using the `docker pause` command ?

Do you have more details on how you'd it manually ? This would be helpful understanding this process.

For now, this is not an available option, but can be added to.

Svenum:
Right I want to use `docker pause` and `docker unpause` instead of `docker stop` and `docker start`

acouvreur:
Can you give me more details on the use case ? Out of curiosity.

Because the memory is not discarded and the process are frozen.

What's the use case that you need to unpause containers to boot very fast ?

valexz:
It appears that the command `docker checkpoint` should be taken into consideration
https://docs.docker.com/engine/reference/commandline/checkpoint/

Svenum:
I do not want to do a snapshot I want to stop the container without the need to start everything on need. I want to reduce CPU Usage and a fast startup. So I do not think that `docker checkpoint` is what we need here.

Svenum:
Some news on this?

acouvreur:
I don't have the time to work on this, feel free to open a pull request if you want :)

PR #618, opened 2025-06-04
Allow pausing and unpausing of containers

chaptergy:
I've had a go at trying to allow pausing and unpausing the containers instead of stopping and starting them. 

This is useful for environments, where the memory usage is not very relevant, but CPU usage and startup time is more important.

Resolves https://github.com/sablierapp/sablier/issues/212.


acouvreur:
Superseded by https://github.com/sablierapp/sablier/pull/755

Issue #650, opened 2025-07-27
Compatible with podman rootless?


Finch290:
**Describe the bug**
I'm using the latest version compatible with podman. 

Trying to reach my dozzle.domain.com pointing on my dozzle container on my VPS, it shows 

> "{
>     "detail": "session was not ready after 1m30s",
>     "instance": "/api/strategies/blocking?group=dozzle\u0026session_duration=2m\u0026timeout=90s",
>     "status": 500,
>     "title": "Internal Server Error",
>     "type": "https://sablierapp.dev/#/errors?id=internal-error"
> }" 

Without sablier as a middleware in traefik, going to dozzle.domain.com let me access dozzle. But as soon as i activate the middleware it shows that error. But the thing is, sablier is waking up dozzle and is stopping it. but it doesn't let me access the URL.

Sablier logs : 

```
8:49PM ERR slog-gin@v1.15.1/middleware.go:265 Error #01: Internal Server Error request.time=2025-07-27T20:48:03.829Z request.method=GET request.host=sablier:10000 request.path=/api/strategies/blocking request.query="group=dozzle&session_duration=2m&timeout=1m" request.params=map[] request.route=/api/strategies/blocking request.ip=10.93.6.85 request.referer="" request.length=0 response.time=2025-07-27T20:49:04.395Z response.latency=1m0.566088993s response.status=500 response.length=272 id=22e16e2c-e0a7-47c6-b14a-30ac604d859e

8:49PM ERR slog-gin@v1.15.1/middleware.go:265 Error #01: Internal Server Error request.time=2025-07-27T20:48:34.689Z request.method=GET request.host=sablier:10000 request.path=/api/strategies/blocking request.query="group=dozzle&session_duration=2m&timeout=1m" request.params=map[] request.route=/api/strategies/blocking request.ip=10.93.6.85 request.referer="" request.length=0 response.time=2025-07-27T20:49:34.741Z response.latency=1m0.051651142s response.status=500 response.length=272 id=c551bbbd-c79b-427b-a8ce-00aff2b98695
```

**Context**
 - Sablier version: 1.10.0
 - Provider: podman 4.3.1
 - Reverse proxy: traefik 3.4.4
 - Sablier running inside a container : yes

my traefik dynamic.yml  : 

```
    sablier:
      plugin:
        sablier:
          sablierUrl: http://sablier:10000
          group: dozzle
          sessionDuration: 2m
          blocking:
            timeout: 1m
```
my sablier compose : 

```
[...]

PR #653, opened 2025-07-31
fix(provider): add debug logging

acouvreur:
Add a bunch of debug logging calls


acouvreur:
Related https://github.com/sablierapp/sablier/issues/650

PR #730, opened 2025-11-17
feat: add nomad provider

### LINKED ISSUE #217: Add `nomad` provider
**Describe the provider**
[Nomad](https://www.nomadproject.io/) is an orchestrator from Hashicorp, similar to Kubernetes but with simpler deployment and operation.

**Does the provider provides the following APIs**
*An instance is a generic term used to describe containers, pods, deployments, and more*

| API | Yes | No | I don't know |
| ---- | ---- | --- | -----------|
| Start an instance |✅| |
| Stop an instance |✅| |
| Get info about an instance |✅| |
| Listening for external events happening  |✅||

**Does the provider has a Go API SDK?**
Yes https://pkg.go.dev/github.com/hashicorp/nomad/api

**Does the provider can be started locally**
Yes

acouvreur:
Closes https://github.com/sablierapp/sablier/issues/217


acouvreur:
Will eventually re-implement it.ac

PR #755, opened 2025-11-27
 feat(docker): add docker pause strategy 

ylecuyer:
Resolves https://github.com/sablierapp/sablier/issues/212

A bit similar to https://github.com/sablierapp/sablier/pull/618

Adds a strategy for the docker provider, either `stop` or `pause` ( `stop` by default for backward compatibility)

Stop strategy (unchanged):

https://github.com/user-attachments/assets/9fce5849-868e-45ff-9256-4c3e7972d1ae

Pause strategy:

https://github.com/user-attachments/assets/72b919fc-3b3f-4b2c-8991-ae428f6c3e6e


acouvreur:
Hello @ylecuyer !

Thanks for the pull request!

I'm not really sure about the strategy parameter. I've seen that using "strategies" often confuse people instead of using some dedicated configuration such as "use-pause".

But I also like it because I might implement the [checkpoint feature](https://docs.docker.com/reference/cli/docker/checkpoint/).

acouvreur:
Nice work @ylecuyer !

I think that's a pretty nice implementation overall! How did you find the testing experience using testcontainers ?

---

I need to enhance the configuration a lot so that it is generated instead.

ylecuyer:
@acouvreur Thanks for merging the PR

> How did you find the testing experience using testcontainers ?

It was good, all transparent I didn't have to worry about anything

PR #763, opened 2025-12-07
better: add latest to docker release

ylecuyer:
Currently the release process doesn't update latest docker tag:

For example current release is 

image

But latest is

image

Which correspond to

image

I added latest tag to go releaser so now latest docker tag will be updated as expected:

image


acouvreur (approved):

acouvreur:
Thanks! I think that I hate :latest tags so much that I didn't think about it :D

But still, people use it, so merging

PR #893, opened 2026-05-10
feat: enforce labels on InstanceStart/InstanceStop operations

louisdtt:
## Summary

This feature follows up issue [#857](https://github.com/sablierapp/sablier/issues/857) and @mzyy94 work, switching `strict-labels` to `ignore-unlabeled`.

It adds a `provider.ignore-unlabeled` flag, disabled by default, to make Sablier ignore resources that are not labeled with `sablier.enable=true`.

When enabled, providers reject start/stop operations for unlabeled instances and filter stopped/remove events so unmanaged resources do not trigger Sablier.

## Changes

- Add `provider.ignore-unlabeled` config and docs.
- Enforce label checks for provider start/stop operations when enabled, keep existing behaviour otherwise.
- Wire the setting into Docker, Docker Swarm, Kubernetes, and Podman providers.
- Filter stopped/remove notifications for unlabeled resources when enabled.
- Add tests for enabled/disabled/labeled/unlabeled behaviour across providers.

## Tests

Ran project tests and currently using this fork with `provider.ignore-unlabeled=true` on a Kind and an AKS cluster.


acouvreur:
Hello @louisdtt 

Since then, we've merged a lot of code, all of it seems out of date unfortunately.

Would you be able to resolve the conflicts ?

louisdtt:
Hello @acouvreur, should be good now 🙂

pkg/sablier/instance_expired.go
```diff
@@ -18,6 +19,10 @@ func OnInstanceExpired(ctx context.Context, provider Provider, recorder metrics.
 			logger.InfoContext(ctx, "instance expired", slog.String("instance", key))
 			err := provider.InstanceStop(ctx, key)
 			if err != nil {
+				if errors.As(err, &ErrInstanceNotManaged{}) {
+					logger.WarnContext(ctx, "instance expired but is not managed by sablier, skipping stop", slog.String("instance", key))
+					return
+				}
```
acouvreur, line 25: I feel like we could avoid all provider code changes and put it at the sablier package level. For example, when we receive an instance expired event, we could get the instance information using InstanceInspect, and check if it is enabled or not.

If it is not enabled, we can log this.

acouvreur (changes requested):
I think we should try to do no code changes in the providers as we expect the same behavior out of them all.

Can we simply call inspect when we receive an instance expiration, and skip the call to InstanceStop if we ignore those ?

Let's park other use cases for now. We might do this in two pull request and two different configurations because they are not the same things:
- One is to fail when a request target resources with instances not tracked by sablier
- The other is to ignore the shutdown for an instance that was (probably) started by sablier but was later removed

louisdtt:
I think you are right, this is a bit much, i'll leave this PR as a draft for now and reopen one focusing on your first point

acouvreur:
Superseded by #908

PR #901, opened 2026-05-13
feat: Reject requests and skip expiration stops for instances without sablier.enable=true

louisdtt:
Adds provider.ignore-unlabeled at Sablier level without changing provider code.
Following #893


docs/configuration.md
```diff
@@ -29,6 +29,8 @@ sablier --configFile=path/to/myconfigfile.yml
 provider:
   # Provider to use to manage containers (docker, swarm, kubernetes, podman, proxmox_lxc)
   name: docker
+  # Reject new requests and skip expiration stops for instances without sablier.enable=true
+  ignore-unlabeled: false
```
acouvreur: Maybe we can change the `ignore-unlabeled` to better express the intent ?

Otherwise everything seems good to me

pkg/sablier/instance_request.go
```diff
@@ -76,9 +76,15 @@ func (s *Sablier) requestStart(ctx context.Context, name string) (InstanceInfo,
 	// the start still proceeds.
 	info, err := s.provider.InstanceInspect(ctx, name)
 	if err != nil {
+		if s.ignoreUnlabeled {
+			return InstanceInfo{}, err
+		}
 		s.l.DebugContext(ctx, "pre-start inspect failed, using bare info", slog.String("instance", name), slog.Any("error", err))
 		info = InstanceInfo{Name: name, CurrentReplicas: 0, DesiredReplicas: 1}
 	}
+	if s.ignoreUnlabeled && info.Enabled != "true" {
```
acouvreur: This would be another flag in my opinion, such as an enforcement feature.

acouvreur (changes requested):
Overall, very nice changes.

But I have the feeling that it should be 2 different flags.

One to require the sablier.enable on the instances
One to ignore expiration on instances that are disabled.

As I mentioned, it is a valid scenario for users to start an instance with the sablier.enable, update the instance to remove sablier.enable and let it run for ever.

Obviously some provider do not support that without restarting the instance, but some do. So I think that it's important.

What do you think ?

I suggest two new flags:
- `require-enable` / `require-sablier-enabled` / other suggestion ? This is for incoming requests.
- `verify-enable-on-expiration` / `stop-on-expiration-if-enabled` / other suggestion ? This is for the expiration function

louisdtt:
Yes, it's better to split the flag, i would say `require-sablier-enabled` for the first one, for the second one `require-sablier-enabled-on-expiration` maybe ? It's hard to write a clear flag that's not too long

acouvreur:
> Yes, it's better to split the flag, i would say `require-sablier-enabled` for the first one, for the second one `require-sablier-enabled-on-expiration` maybe ? It's hard to write a clear flag that's not too long

I think those express a better intent:

`provider.reject-unlabeled-requests`

Purpose: reject incoming requests when the target instance is not sablier.enable=true.

EDIT: It's worth noting that `provider.reject-unlabeled-requests` is only intended for the API using the `names` query param, and not the `group`. Because by definition, all groups only retrieve enabled instances.

`provider.verify-enabled-on-expiration`

Purpose: on expiration, re-check the instance label and only stop if sablier.enable=true.

I've made Claude Sonnet analyze the PR and give some suggestion, and I think I agree with those suggestions, the intent is very clear and represent properly the new features

acouvreur:
@louisdtt I'm sorry I've introduced new command flags recently, you have to rebase again. I will wait for your rebase before merging any further PRs

acouvreur:
> @louisdtt I'm sorry I've introduced new command flags recently, you have to rebase again. I will wait for your rebase before merging any further PRs

Ok I lied, I've just merged https://github.com/sablierapp/sablier/pull/907

Let me know if you need help to resolve the merge conflict

louisdtt:
Don't worry, l'll do it in a few hours

acouvreur:
All tests ok, the build error is just sonar

pkg/sablier/session_request_test.go
```diff
@@ -127,6 +128,41 @@ func TestSessionsManager(t *testing.T) {
 	})
 }
 
+func TestRequestSessionGroup_DoesNotRejectUnlabeledInstances(t *testing.T) {
```
acouvreur, line 155: Shouldn't we have the same test for a rejection ?

pkg/sablier/instance_expired.go
```diff
@@ -13,9 +13,28 @@ import (
...
+}
+
+func onInstanceExpired(ctx context.Context, provider Provider, recorder metrics.Recorder, logger *slog.Logger, verifyEnabled bool) func(string) {
 	return func(_key string) {
 		go func(key string) {
 			logger.InfoContext(ctx, "instance expired", slog.String("instance", key))
+			if verifyEnabled {
+				info, err := provider.InstanceInspect(ctx, key)
+				if err != nil {
+					logger.ErrorContext(ctx, "instance expired could not be inspected before stop", slog.String("instance", key), slog.Any("error", err))
```
acouvreur: Could be a warning instead

acouvreur (dismissed):
LGTM

A few code style could be improved, but we can do it in another pull request

acouvreur:
Maybe one last thing, @louisdtt could you create 2 working examples under examples/ so that it is easy to demonstrate and test in an end to end scenario ? After that, we can merge if testing is successful

pkg/sablier/session_request_test.go
louisdtt: done

acouvreur:
Ok, let me try the examples and we're good to go

acouvreur:
Ok, I've tried both example. Both works, but I had a few issues with the verify example:

```
make test
docker compose up -d
[+] Running 3/3
 ✔ Network verify-enabled-on-expiration_default      Created                                                                                                                                                                     0.0s
 ✔ Container sablier-verify-expiration-unlabeled     Started                                                                                                                                                                     0.5s
 ✔ Container verify-enabled-on-expiration-sablier-1  Started                                                                                                                                                                     0.4s
curl: (56) Recv failure: Connection reset by peer
==> Stopping unlabeled container before requesting a session
docker compose stop unlabeled
[+] Stopping 1/1
 ✔ Container sablier-verify-expiration-unlabeled  Stopped                                                                                                                                                                        0.1s
==> Requesting unlabeled container by names
curl -fsS 'http://localhost:10000/api/strategies/blocking?names=sablier-verify-expiration-unlabeled&session_duration=3s&timeout=10s' > "/tmp/sablier-verify-enabled-on-expiration-response.json"
curl: (22) The requested URL returned error: 500
make: *** [test] Error 22
```

I will merge it now as this is just an example.

acouvreur (approved):

PR #911, opened 2026-05-14
fix: example verify enabled on expiration

louisdtt:
Change strategy to dynamic

Edit: Fixed delay and keep blocking instead


examples/verify-enabled-on-expiration/Makefile
acouvreur, line 3: If you are switching to dynamic, it won't be json

examples/verify-enabled-on-expiration/Makefile
```diff
@@ -26,7 +26,7 @@ test: up wait-sablier
 	@echo "==> Stopping unlabeled container before requesting a session"
 	docker compose stop unlabeled
 	@echo "==> Requesting unlabeled container by names"
-	curl -fsS '$(SABLIER_URL)/api/strategies/blocking?names=$(UNLABELED_NAME)&session_duration=3s&timeout=10s' > "$(RESPONSE_FILE)"
+	curl -fsS '$(SABLIER_URL)/api/strategies/dynamic?names=$(UNLABELED_NAME)&session_duration=3s' > "$(RESPONSE_FILE)"
```
acouvreur: I think blocking is the right strategy, however, the API might have an issue

acouvreur (approved):

PR #920, opened 2026-05-16
feat: add webhooks

### LINKED ISSUE #423: Add webhooks for `started` and `stopped` events
Add webhooks for `started` and `stopped` events.

A webhook can be defined as:
- an URL
- an optional custom payload

More customization may be added later.

---

The first goal is to be able to create such integration: https://github.com/louislam/uptime-kuma/issues/5228

Instead of Uptime-Kuma adding specific code for Sablier, we will use these webhooks to call the Uptime Kuma API.

acouvreur:
You can now configure webhooks that will be called upon a specific event (started or stopped)

Closes #423

### DIFF
--- a/pkg/webhook/dispatcher.go
+++ b/pkg/webhook/dispatcher.go
@@ -1,0 +1,184 @@
+// Package webhook delivers normalized HTTP notifications when Sablier-managed
+// instances change state. It acts as an abstract, provider-agnostic event
+// bridge: regardless of whether the underlying runtime is Docker, Kubernetes,
+// Docker Swarm, Podman, or Proxmox LXC, every consumer receives the same JSON
+// payload structure.
+package webhook
+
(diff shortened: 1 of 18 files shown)


pkg/webhook/dispatcher.go
```diff
@@ -0,0 +1,184 @@
...
+	for k, v := range ep.Headers {
+		req.Header.Set(k, v)
+	}
+
+	resp, err := d.client.Do(req)
+	if err != nil {
+		d.l.ErrorContext(ctx, "webhook: delivery failed", slog.String("url", ep.URL), slog.Any("error", err))
+		return
+	}
+	defer resp.Body.Close()
```
acouvreur:
```suggestion
	defer resp.Body.Close() //nolint:errcheck
```

PR #944, opened 2026-05-27
feat(provider): support CloudNativePG Cluster hibernation in the Kubernetes provider

### LINKED ISSUE #943: feat: support CloudNativePG Cluster (hibernate) in the Kubernetes provider
**Describe the provider**

This is not strictly a new provider, but an extension of the existing Kubernetes provider to handle CloudNativePG (CNPG) `Cluster` resources (`postgresql.cnpg.io/v1`) as a new `kind` alongside `deployment` and `statefulset`.

CNPG ships with a built-in declarative hibernation mechanism, triggered by an annotation on the `Cluster` CR:

- `cnpg.io/hibernation: on` -> the operator scales the workload down, removes the underlying StatefulSet, but keeps the PVCs intact.
- `cnpg.io/hibernation: off` (or annotation removed) -> the operator rebuilds the StatefulSet and the cluster comes back online.

This maps very cleanly onto Sablier's `InstanceStop` / `InstanceStart` semantics, so a small extension to the Kubernetes provider would unlock CNPG-backed workloads.

**Why this matters (use case)**

Our typical workload (Opencell) is a stack of three pieces deployed in the same cluster:

- an Opencell `StatefulSet`
- a Keycloak `Deployment`
- a CNPG `Cluster` for the database

We'd like to put all three in the same `sablier.group=opencell` so that hitting the Traefik ingress wakes the whole stack up, and inactivity hibernates the whole stack. Today the first two are covered out of the box; only the CNPG `Cluster` is missing. Note that scaling the operator-managed StatefulSet directly is not viable (the operator reconciles it back), which is why going through the official hibernation annotation is the right path.

**Does the provider provide the following APIs?**

| API                                     | Yes | No | I don't know |
| --------------------------------------- | --- | -- | ------------ |
| Start an instance                       |  x  |    |              |
| Stop an instance                        |  x  |    |              |
| Get info about an instance              |  x  |    |              |
| Listening for external events happening |  x  |    |              |

- Start: `PATCH` the `Cluster` CR to set `cnpg.io/hibernation: off` (or remove the annotation).
- Stop: `PATCH` the `Cluster` CR to set `cnpg.io/hibernation: on`.
- Inspect: read `spec.instances`, `status.readyInstances` and the hibernation annotation to compute `Ready` / `Starting` / `Stopped`.
- Events: dynamic informer on `clusters.postgresql.cnpg.io` to emit the `stopped` event when hibernation transitions to `on`.

**Does the provider have a Go API SDK?**

Two options, both viable:

- Typed client: https://github.com/cloudnative-pg/cloudnative-pg (`api/v1` package). Cleanest, but adds a non-trivial dependency (pulls controller-runtime transitively).
- Dynamic client from `client-go` (`dynamic.Interface`). Already available transitively in Sablier, no new dependency. A bit less ergonomic but perfectly sufficient for the surface we need (annotation patch + read a couple of status fields + informer).

Our preference would be the dynamic client to keep the dependency footprint minimal, but happy to align with whatever the maintainers prefer.

**Can the provider be started locally?**

Yes. CNPG is installable in any local cluster (kind, k3s, minikube) via a single manifest:

```bash
kubectl apply --server-side -f \
  https://raw.githubusercontent.com/cloudnative-pg/cloudnative-pg/main/releases/cnpg-1.x.y.yaml
```

This makes it easy to wire up testcontainers-based integration tests in the same spirit as the existing Kubernetes provider tests.

**Proposed shape of the change**

Mostly additive, following the existing per-kind file pattern in `pkg/provider/kubernetes/`:

- New files: `cnpgcluster_inspect.go`, `cnpgcluster_list.go`, `cnpgcluster_events.go`, `cnpgcluster_hibernate.go` (+ tests).
- Small edits to route the new kind in `parse_name.go`, `instance_inspect.go`, `instance_start.go`, `instance_stop.go`, `instance_list.go`, `instance_events.go`, `kubernetes.go`.
- New doc page `docs/providers/cnpg.md` (sibling of the existing provider docs).
- No change to the `sablier.Provider` interface, no new CLI flag required.

Rough estimate: ~400-600 LOC including tests.

**Additional context**

If this is something you'd be open to merging, we'd be glad to do the implementation work and submit a PR. We just wanted to surface the proposal and the design choices first to avoid building something that wouldn't be welcome.

Happy to discuss the design (typed client vs dynamic client, naming of the kind, readiness semantics for multi-instance clusters, etc.) before writing any code.

Thanks for the project!

antoinemichea:
Closes #943

## What this does

Following up on the discussion in #943, this adds a new `cnpgcluster` workload kind to the Kubernetes provider so Sablier can start and stop [CloudNativePG](https://cloudnative-pg.io/) `Cluster` resources, right next to Deployments and StatefulSets.

Instead of scaling a replica count, Sablier toggles CloudNativePG's declarative hibernation annotation:

- **stop** → `cnpg.io/hibernation: "on"` (operator scales the cluster down and removes its workload, **keeps the PVCs**)
- **start** → `cnpg.io/hibernation: "off"` (operator resumes the cluster)

My use case: I run Opencell as a StatefulSet + a Keycloak + a CloudNativePG database. With this, I can put all three in the same `sablier.group`, so a single request on the Traefik ingress wakes the whole stack and inactivity hibernates all of it — database included.

## Design choices

- **Dynamic client, no new dependency.** I went with `k8s.io/client-go/dynamic` (already pulled by client-go) rather than the typed CloudNativePG client, to keep the dependency footprint minimal — no `controller-runtime`. The dynamic client is wired into the provider next to the typed clientset.
- **Purely additive.** No change to the `sablier.Provider` interface, no new CLI flag. The new kind is routed inside the existing `instance_{start,stop,inspect,list,events}.go`.
- **Safe when CloudNativePG is absent.** If the CRD isn't installed, listing no-ops (returns no clusters instead of erroring) and the events informer isn't started, so non-CNPG clusters are completely unaffected.
- **Readiness mapping:** hibernation `on` → stopped; `status.readyInstances >= spec.instances` → ready; otherwise starting.

New files live alongside the existing per-kind files: `cnpgcluster.go`, `cnpgcluster_hibernate.go`, `cnpgcluster_inspect.go`, `cnpgcluster_list.go`, `cnpgcluster_events.go`, plus tests. Docs updated in `docs/providers/kubernetes.md` (new section + the optional RBAC rule).

## Testing done

I went a bit further than usual on testing because it touches a database:

- **Unit tests** (`cnpgcluster_test.go`) using `client-go/dynamic/fake`: status mapping (ready/starting/stopped), hibernate patching via `InstanceStart`/`InstanceStop`, and list/groups label filtering.
- **Testcontainers integration test** (`cnpgcluster_integration_test.go`), in the same spirit as the existing k3s tests: it installs a minimal Cluster CRD into the shared k3s cluster and verifies inspect / stop / start / list against the real API server. As you mentioned in #943, the CNPG install is done in the test itself (CRD only — no operator needed to validate Sablier's interactions).
- **Real cluster, provider level:** drove the provider against a real CloudNativePG cluster and watched the actual operator hibernate (pods → 0, PVC kept) and resume.
- **Real cluster, full end-to-end:** deployed this build in-cluster behind a Traefik instance with the Sablier plugin, with a CloudNativePG `Cluster` and a `whoami` app sharing one `sablier.group`. First request wakes the whole group (app scaled up + database resumed), then after the session expires both are stopped together (deployment → 0, cluster → hibernated), and a later request blocks ~25s while the database recovers, then serves. Exactly the behaviour I was after.
- `make fmt`, `golangci-lint run`, and the full `pkg/provider/kubernetes` test suite are green.

Happy to adjust anything — naming of the kind, readiness semantics for multi-instance clusters, etc.

### DIFF
--- a/docs/providers/kubernetes.md
+++ b/docs/providers/kubernetes.md
@@ -59,7 +59,19 @@ rules:
       - get     # Retrieve info about specific dep
       - list    # Events
       - watch   # Events
+  # Only required if you manage CloudNativePG Clusters (see below).
+  - apiGroups:
+      - postgresql.cnpg.io
+    resources:
+      - clusters
+    verbs:
+      - get     # Retrieve info about a specific cluster
+      - list    # Discovery and events
+      - watch   # Events
+      - patch   # Toggle the hibernation annotation
 ```
+
+?> The `postgresql.cnpg.io` rule is optional. When the CloudNativePG CRD is absent, Sablier simply skips CloudNativePG discovery.
 
 ## Register Deployments
 
@@ -96,3 +108,38 @@ spec:
 Sablier checks for the deployment replicas. As soon as the current replicas matches the wanted replicas, then the deployment is considered `ready`.
 
 ?> Kubernetes uses the Pod healthcheck to check if the Pod is up and running. So the provider has a native healthcheck support.
+
+## Register CloudNativePG Clusters
+
+Sablier can also start and stop [CloudNativePG](https://cloudnative-pg.io/) `Cluster` resources. Instead of scaling a replica count, Sablier toggles CloudNativePG's [declarative hibernation](https://cloudnative-pg.io/documentation/current/declarative_hibernation/) annotation:
+
+- **Stop** sets `cnpg.io/hibernation: "on"` — the operator scales the cluster down and removes its workload while keeping the PVCs.
+- **Start** sets `cnpg.io/hibernation: "off"` — the operator resumes the cluster.
+
+Opt-in with the same labels used for deployments and statefulsets:
+
+```yaml
+apiVersion: postgresql.cnpg.io/v1
+kind: Cluster
+metadata:
+  name: opencell-db
+  labels:
+    sablier.enable: "true"
+    sablier.group: opencell
+spec:
+  instances: 3
+  storage:
+    size: 1Gi
+```
+
+This makes it possible to put an application, its Keycloak and its database in a single `sablier.group`, so that a single request wakes up the whole stack and inactivity hibernates all of it.
+
+### Cluster readiness
+
+A CloudNativePG Cluster is considered:
+
+- `stopped` when the `cnpg.io/hibernation` annotation is `"on"`;
+- `ready` when `status.readyInstances` is greater than or equal to `spec.instances`;
+- `starting` otherwise.
+
+?> Resuming a hibernated cluster takes longer than a simple scale-up (PVC reattachment and PostgreSQL recovery). Make sure your reverse-proxy timeouts allow for it.
--- a/pkg/provider/kubernetes/cnpgcluster.go
+++ b/pkg/provider/kubernetes/cnpgcluster.go
@@ -1,0 +1,44 @@
+package kubernetes
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// KindCNPGCluster is the workload kind used in instance names to identify a
+// CloudNativePG Cluster custom resource (e.g. "cnpgcluster_namespace_name_1").
+const KindCNPGCluster = "cnpgcluster"
+
+// cnpgClusterGVR is the GroupVersionResource of CloudNativePG Cluster resources.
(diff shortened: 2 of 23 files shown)


acouvreur (approved):
LGTM!

Thank you for this feature, I hope this comes handy to you :)

PR #956, opened 2026-05-31
feat(docker): respect restart policy and resolve `depends_on` on start

### LINKED ISSUE #792: Correct way of setting up a multi service compose stacks with `depends_on` and traefik labels
Is there a correct way of setting up multi service compose stacks with dependencies?
For example the classic one: an app with a db where the app needs to wait until the db is healthy: 
```yaml
app:
    depends_on:
      postgresql:
        condition: service_healthy
    labels:
      - traefik.http.routers.${STACK_NAME}.middlewares=${STACK_NAME}-sablier
      - traefik.http.middlewares.${STACK_NAME}-sablier.plugin.sablier.group=${STACK_NAME}
      - traefik.http.middlewares.${STACK_NAME}-sablier.plugin.sablier.sablierUrl=http://sablier-app-1:10000
      - traefik.http.middlewares.${STACK_NAME}-sablier.plugin.sablier.sessionDuration=1m
      - traefik.http.middlewares.${STACK_NAME}-sablier.plugin.sablier.dynamic.displayName=${STACK_NAME}
      - sablier.enable=true
      - sablier.group=${STACK_NAME}
      - traefik.docker.allownonrunning=true
postgresql:
    labels:
      - sablier.enable=true
      - sablier.group=${STACK_NAME}
```

My setup basically works as expected. _sablier_ is shutting both down and starts them when I open the url. But it starts both containers at the same time which results in errors in the app because the db is not yet ready.

### LINKED ISSUE #952: Sablier should restart a container only once
**Describe the bug**
Sablier keeps restarting a container after it ends its function if other containers in the same group aren't ready.

**Context**
 - Sablier version: 1.12.0
 - Provider: Podman 5.4.2
 - Reverse proxy: Caddy 2.11.3
 - Sablier running inside a container? Yes

**Expected behavior**
Sablier should restart a container just once and leave the restarting resolution to the docker configuration (Restart Policies: no, always, on-failure, unless-stopped) 

**Additional context**
[Affine](https://docs.affine.pro/self-host-affine/install/docker-compose-recommend) uses 3 containers. The main service, the worker (affine-indexer), and the pre-start migration (affine-migration). 

The web server container must wait for the migration container to end before it continues to boot (depends_on: affine-migration: condition: service_completed_successfully).
As a result, sablier keeps unnecessarily restarting  the migration container. That causes the affine server never to be ready, because it is always waiting for its end. 
If I remove the _depends_on_ condition, the affine server will start prematurely, before the migration container finishes.
If I set the sablier.enable label to false, the pre-migration container will never do its pre-startup work.

**docker compose:**
```
%YAML 1.1
---
x-service:
  environment: &defenv
    AFFINE_SERVER_EXTERNAL_URL: "https://affine.domain.pro"
    AFFINE_INDEXER_ENABLED: true
    AFFINE_INDEXER_SEARCH_ENDPOINT: "http://affine-indexer:9308"
    LOG_LEVEL: warn
services:
  affine: ##########################################
    depends_on:
      affine-migration:
        condition: service_completed_successfully
    image: ghcr.io/toeverything/affine:stable
    container_name: affine
    environment:
      <<: *defenv
    volumes:
      - /xxx/storage:/root/.affine/storage
      - /xxx/config:/root/.affine/config
    networks: [affinenet]
    ports: ["46005:3010"]
    healthcheck: 
      test: ["CMD", "node", "-e", "require('net').connect(3010,'localhost',()=>process.exit(0)).on('error',()=>process.exit(1)).setTimeout(2000,()=>process.exit(1))"]
      interval: 30s
      timeout: 5s
      retries: 5
      start_period: 6s
    labels:
      - sablier.enable=true
      - sablier.group=Affine
  affine-migration: ##########################################
    image: ghcr.io/toeverything/affine:stable
    container_name: affine-migration
    command: ['sh', '-c', 'node ./scripts/self-host-predeploy.js']
    environment:
      <<: *defenv
    networks: [affinenet]
    volumes:
      - /xxx/storage:/root/.affine/storage
      - /xxx/config:/root/.affine/config
    healthcheck:
      test: ["CMD-SHELL", "echo 'Healthy'"]
      interval: 8s
      timeout: 1s
      retries: 3
      start_period: 2s
    labels:
      - sablier.enable=true
      - sablier.group=Affine
  affine-indexer: ##########################################
    ulimits:
      nproc: 65535
      nofile:
        soft: 65535
        hard: 65535
      memlock:
        soft: -1
        hard: -1
    image: docker.io/manticoresearch/manticore:10.1.0
    container_name: affine-indexer
    environment:
      <<: *defenv
    networks: [affinenet]
    volumes:
      - /xxx/manticore:/var/lib/manticore
    healthcheck:
      test: ['CMD', 'wget', '-O-', 'http://127.0.0.1:9308']
      interval: 30s
      timeout: 5s
      retries: 5
      start_period: 5s
    labels:
      - sablier.enable=true
      - sablier.group=Affine
...
```

acouvreur:
This PR addresses two closely related issues that prevented Sablier from working correctly with real-world multi-service Compose stacks: init/migration containers and ordered startup with `depends_on`.

Closes #792
Closes #952

---

### 1. Respect the Docker restart policy for exited containers (`#952`)

**The problem.** When a container finishes its job and exits with code `0` (a common pattern for one-shot init or schema-migration containers), Sablier was reporting it as `Stopped`. This caused Sablier to restart the container on every readiness poll, which in turn prevented the main application from ever becoming ready — because it was always waiting for its `service_completed_successfully` dependency to finish.

**The fix.** `InstanceInspect` now consults the container's restart policy before mapping an exit-0 state:

| Restart policy | Exit 0 → Sablier status | Why |
|---|---|---|
| `always` / `unless-stopped` | `Starting` | Docker will restart the container; the exit is transient |
| `no` / `on-failure` | **`Ready` (completed)** | The container completed its work and won't be restarted |

Non-zero exits are still reported as `Error` regardless of policy, so real failures are not masked.

```go
// pkg/provider/docker/container_inspect.go
case container.StateExited:
    if spec.Container.State.ExitCode != 0 {
        return errorStatus(...)
    }
    if restartsOnSuccess(restartPolicyMode(spec.Container.HostConfig)) {
        return startingStatus(...) // transient — Docker will bring it back
    }
    return readyStatus(1, ...) // completed one-shot / init container ✅
```

---

### 2. Resolve `depends_on` before starting a container (`#792`)

**The problem.** Sablier starts every container in a group concurrently. When a group contains both an application and its database (or a migration), the app often starts before the database is ready, leading to connection errors.

**The fix.** `InstanceStart` now reads the `com.docker.compose.depends_on` label that Docker Compose writes automatically and resolves the full dependency graph before starting the target container:

```
Request to start "app"
  └─ reads depends_on label: db:service_healthy, migration:service_completed_successfully
       ├─ start "db" → wait until healthy ✅
       └─ start "migration" → wait until exited 0 ✅
  └─ start "app" ✅
```

All four Docker Compose conditions are supported:

| Condition | Behaviour |
|---|---|
| `service_started` | Waits until `State.Running == true` |
| `service_healthy` | Waits until the health check passes |
| `service_completed_successfully` | Waits until `StateExited` + exit code 0 |
| `service_running_or_healthy` | Running, or healthy if a health check exists |

Dependencies that are not in the same Compose project (or have no matching container) are skipped with a warning — Sablier never fails a start because of an optional or external dependency.

Cycle protection is built in: a `started` set is threaded through every recursive `instanceStart` call; any container that is already being started is skipped.

> **No label changes needed on your existing stack.** The `depends_on` label is written by `docker compose up` automatically. You only need `sablier.enable` / `sablier.group` on the container you want Sablier to manage.

---

## How to migrate

No configuration changes are required. The behaviour changes automatically:

**Before this PR** — Affine / Paperless / any stack with a migration container:
```
Sablier starts affine-migration ──► exits 0 ──► Sablier reports Stopped
                                                 ──► Sablier restarts it
                                                      ──► affine never starts 🔴
```

**After this PR:**
```
Sablier starts affine-migration ──► exits 0, restart:no ──► Ready (completed)
Sablier starts affine ──────────────────────────────────────────────────────► 🟢
```

---

## Runnable example

A complete `docker compose` stack is provided under [`examples/depends-on/`](./examples/depends-on/):

```
services:
  sablier   — manages the "app" group
  db        — mimic with health check (service_healthy dependency)
  migration — one-shot mimic (exit 0, restart:no → service_completed_successfully)
  app       — the managed service; depends_on db (healthy) + migration (completed)
```

```bash
cd examples/depends-on
make up     # start sablier, stop the dependency chain
make start  # blocking request — watch Sablier resolve the graph
make down
```

---

## Tests

- **Unit:** `TestParseComposeDependsOn` — table-driven tests for the label parser (no daemon required, runs in `-short` mode).
- **Integration:** `TestDockerClassicProvider_StartWithDependsOn` and `TestDockerClassicProvider_StartWithDependsOnHealthy` — dind-based tests that create real containers, set `com.docker.compose.*` labels, call `InstanceStart`, and assert the full dependency chain was started in the correct order.
- **Inspect regression:** The existing `exited container with status code 0` test has been updated to assert `Ready` (was `Stopped`); a new `on-failure` restart-policy variant is added.

---

EDIT: Not satisfied with current implementation

Will have to properly implement in two phases:
- Startup dependency at the sablier level, not inside the provider. But each provider can give sablier hints about the dependencies
- What is considered "ready" -> respecting restart policies


antoinemichea:
Heads-up: `main` is currently red since this merged (Build shows ~39 failures). Looks like two separate things:

1. **Stale inspect test expectations after the health-status change.** This PR changed a running-but-unhealthy container to report `starting` / "container is running but not healthy yet" (instead of `error` / "container is unhealthy") and updated the docker inspect tests accordingly — but some expectations still assert the old value. For example the podman case `running container with "unhealthy" health` in `pkg/provider/podman/container_inspect_test.go` still expects `Status: InstanceStatusError` / "container is unhealthy". Since podman shares the docker status derivation, it now returns `starting` and the test fails. Just needs the `want` aligned with the new behaviour.
2. **kubernetes failures look unrelated / flaky.** The `kubernetes` inspect/start/stop tests fail with `connection refused` to the k3s testcontainer API (`Get "https://localhost:32769/version"`), i.e. the k3s container wasn't reachable — likely CI flakiness that passes on a re-run rather than a code issue.

Happy to open a small PR aligning the podman inspect test expectations if that helps.

PR #963, opened 2026-06-04
feat(provider): pause redis-operator reconciliation during StatefulSet scale-to-zero

### LINKED ISSUE #962: OT-CONTAINER-KIT redis-operator reconciles StatefulSet replicas back to 1 after Sablier scales to zero
## What's happening

I have several homelab applications sitting behind Traefik with Sablier managing scale-to-zero. After enabling the recent CloudNativePG support I tried to extend the same pattern to Redis instances managed by the [OT-CONTAINER-KIT redis-operator](https://github.com/OT-CONTAINER-KIT/redis-operator).

Sablier correctly discovers the StatefulSet (via the `sablier.enable: "true"` label, which the operator propagates from the Redis CR to the StatefulSet it creates), and scales it to zero on session expiry. But the operator's reconciliation loop sees the replica count drop and immediately restores it to 1 — so the pod never actually stops.

## Why it happens

The redis-operator continuously reconciles its managed StatefulSets back to the desired replica count. Unlike CloudNativePG, which Sablier can pause via the hibernation annotation, there's no built-in mechanism for Sablier to tell the operator to stand down.

The operator does ship a pause mechanism: setting `redis.opstreelabs.in/skip-reconcile: "true"` on the Redis CR causes the controller to skip its reconciliation loop for that resource. Setting this annotation before scaling to zero, and removing it after scale-up, is enough to make scale-to-zero work reliably.

## Environment

- redis-operator: OT-CONTAINER-KIT v0.24.0 (`redis.redis.opstreelabs.in/v1beta2`)
- Sablier: 1.13.0 (Kubernetes provider)
- Observed with standalone `Redis` kind; `RedisCluster`/`RedisReplication`/`RedisSentinel` are out of scope for this issue

jlaska:
Closes #962

## What this does

I run a homelab where several applications sit behind Traefik with Sablier managing scale-to-zero. After the recent CloudNativePG support landed I started enabling it across my stack, and noticed Redis instances managed by the OT-CONTAINER-KIT redis-operator were the odd one out — Sablier would scale the StatefulSet to zero on session expiry, but the operator would immediately reconcile replicas back to 1.

The operator already ships a pause mechanism: `redis.opstreelabs.in/skip-reconcile: "true"` on the Redis CR. Setting this before scaling to zero keeps the operator out of the way, and clearing it after scale-up hands control back.

- **stop** → set `skip-reconcile: "true"` on the owning Redis CR, then scale the StatefulSet to 0
- **start** → scale the StatefulSet back to 1, then remove the annotation so the operator resumes normal reconciliation
- If the scale fails, a cleanup defer removes the annotation so the operator is never left paused with pods still running.

## Design choices

- **No new kind.** This hooks into the existing StatefulSet path rather than introducing a `redis` kind. The operator propagates labels (including `sablier.enable`/`sablier.group`) from the Redis CR to the StatefulSet it creates, so existing label-based opt-in works without any change to how users configure Sablier.
- **Only on scale-to-zero.** The annotation is not set in scale-mode stops (`sablier.idle.replicas >= 1`), where the StatefulSet stays non-zero and the operator should continue reconciling normally.
- **Version-agnostic owner detection.** The owner reference check matches on API group only (`redis.redis.opstreelabs.in`), not on the specific version string, so it stays correct if the operator promotes to v1.
- **Best-effort.** If the dynamic client is not configured or the Redis CR cannot be fetched, a warning is logged and the scale proceeds unchanged — existing behaviour for non-redis-operator StatefulSets is completely unaffected.

## Testing done

- Unit tests (`statefulset_redis_operator_test.go`) covering owner detection (including forward-compatibility with a future v1 APIVersion), annotation set/clear, no-op for plain StatefulSets, stop annotation set with cleanup-on-failure, and start annotation removal.
- `make fmt`, `golangci-lint run`, and the full `pkg/provider/kubernetes` test suite pass.
- Real cluster end-to-end: session expiry scales the StatefulSet to 0 with `skip-reconcile: "true"` set; the operator logs the annotation and stands down. A new request scales back to 1 and the annotation is removed. The redis-operator resumes normally.

### DIFF
--- a/pkg/provider/kubernetes/statefulset_redis_operator_test.go
+++ b/pkg/provider/kubernetes/statefulset_redis_operator_test.go
@@ -1,0 +1,314 @@
+package kubernetes
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/neilotoole/slogt"
+	"go.opentelemetry.io/otel"
+	"gotest.tools/v3/assert"
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	dynamicfake "k8s.io/client-go/dynamic/fake"
+	k8sfake "k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+)
+
+// newRedisOwnerSTS builds a StatefulSet whose controller owner is a Redis CR.
+func newRedisOwnerSTS(namespace, name string, replicas int32, apiVersion string) *appsv1.StatefulSet {
+	if apiVersion == "" {
+		apiVersion = redisOperatorAPIVersion()
+	}
+	isController := true
+	return &appsv1.StatefulSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    map[string]string{"sablier.enable": "true"},
+			OwnerReferences: []metav1.OwnerReference{{
+				APIVersion: apiVersion,
+				Kind:       "Redis",
+				Name:       name, // Redis CR has same name as the StatefulSet by convention
+				Controller: &isController,
+			}},
+		},
+		Spec: appsv1.StatefulSetSpec{
+			Replicas: &replicas,
+		},
+		Status: appsv1.StatefulSetStatus{ReadyReplicas: replicas},
+	}
+}
+
+// redisOperatorAPIVersion returns the canonical APIVersion string for the operator.
+func redisOperatorAPIVersion() string {
+	return redisOperatorGVR.Group + "/v1beta2"
+}
+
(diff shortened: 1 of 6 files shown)


pkg/provider/kubernetes/statefulset_redis_operator_test.go
acouvreur, line 1: Would you be able to add a test using the same testcontainer mechanism which actually deploys a redis operator and make sure this works as expected for your case ?

You can keep theses tests, but I'm more interested in integration tests here, not api mocking

That way I can safely know if this will work for real.

I've caught many bugs switching from mocks to testcontainers here.

jlaska:
@acouvreur I added `statefulset_redis_operator_integration_test.go` following the same pattern as the CNPG test — it installs a minimal Redis CRD into the shared k3s cluster and creates a Redis CR alongside a StatefulSet whose `ownerReference` points to it, simulating what the operator produces.

The three sub-tests verify Sablier's behavior against a real API server:

- **stop**: `InstanceStop` sets `skip-reconcile: "true"` on the Redis CR and scales the StatefulSet to 0
- **inspect**: `InstanceInspect` correctly reports the stopped state
- **start**: `InstanceStart` scales back to 1 and clears the annotation

No operator binary is deployed — the tests validate Sablier's own API interactions with real Kubernetes objects, not operator behavior.

Let me know if there are additional changes you'd like.

acouvreur (approved):
LGTM

Thank you for the pull request

PR #980, opened 2026-06-27
Start session on group discovery

louisdtt:
Adds optional session seeding with flag `--sessions.create-on-group-discovery`.

When `--sessions.create-on-group-discovery=true` is set, Sablier creates a `sessions.default-duration` session for newly discovered instances. 

This makes sure that a session is attached to a group and workloads expire and downscale without waiting for a first request to start the session.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -251,6 +251,8 @@ storage:
 sessions:
   # The default session duration (default 5m)
   default-duration: 5m
+  # Create a default-duration session when a group-managed instance is discovered (default false)
+  create-on-group-discovery: false
   # The expiration checking interval. 
   # Higher duration gives less stress on CPU. 
   # If you only use sessions of 1h, setting this to 5m is a good trade-off.
@@ -358,6 +360,7 @@ sablier start --strategy.dynamic.custom-themes-path /my/path
       --server.base-path string                               The base path for the API (default "/")
       --server.metrics.enabled                                Enable the Prometheus /metrics endpoint (default false)
       --server.port int                                       The server port to use (default 10000)
+      --sessions.create-on-group-discovery                    Create a default-duration session when a group-managed instance is discovered
       --sessions.default-duration duration                    The default session duration (default 5m0s)
       --sessions.expiration-interval duration                 The expiration checking interval. Higher duration gives less stress on CPU. If you only use sessions of 1h, setting this to 5m is a good trade-off. (default 20s)
       --storage.file string                                   File path to save the state
--- a/examples/group-discovery-session/Makefile
+++ b/examples/group-discovery-session/Makefile
@@ -1,0 +1,29 @@
+SABLIER_URL ?= http://localhost:10000
+WHOAMI_NAME := sablier-group-discovery-whoami
+
+.PHONY: up down logs wait-sablier test
+
+up:
+	docker compose up -d
+
+down:
+	docker compose down
+
+logs:
+	docker compose logs -f sablier
+
+wait-sablier:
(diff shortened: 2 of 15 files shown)


acouvreur:
I think `--provider.auto-stop-externally-started` should do that automatically already

louisdtt:
I did not think of this option, but this would mean that new deployments are scaled to 0 by default right ? The idea behind this is more to start them as usual and make sure they don't hang forever

acouvreur:
> I did not think of this option, but this would mean that new deployments are scaled to 0 by default right ? The idea behind this is more to start them as usual and make sure they don't hang forever

Hey @louisdtt thanks again for the pull request. Can you take a look at what's been discussed in https://github.com/sablierapp/sablier/issues/985 ? I think the main purpose is the same, but I'm more in favor of what's been proposed there. 

What do you think ?

antoinemichea:
Hi @louisdtt, thanks for the PR, and for being looped in here (context: #985). We're after the exact same goal: make sure a running, group-managed workload gets a session so it expires on its own instead of hanging forever. Your implementation already gets the tricky parts right: it skips when a session already exists (`sessions.Get`), only seeds `Ready` instances, and seeds at startup via `SeedGroupSessions`, so no renewal loop.

On @acouvreur's suggestion of deploying with 0 replicas: that isn't viable for us. Our environments must be up right after deployment, because our deploy pipeline and clients issue API calls to the app immediately (schema/data provisioning, readiness checks, automated tests), and some post-deploy steps require the app to be running. Deploying at 0 replicas would send those calls to a scaled-to-zero backend (failures / cold starts) and break the pipeline. So we deploy at 1 and register the Sablier session a moment later via the warm API, which is exactly what creates the "running but not yet sessioned" window.

On convergence: our proposal in #985 is essentially the same behavior, framed as a symmetric counterpart to `auto-stop-externally-started`: same detection (enabled + running + no session), opposite action (create a session instead of stopping). It overlaps heavily with what you've built; the main differences are placement (a dedicated externally-started watcher vs. seeding inside `group_watch`) and naming. @acouvreur mentioned he leans toward the symmetric-option framing.

Rather than open a competing PR, we'd much rather converge on a single solution with you. Two points to align:

1. Framing/placement: keep `sessions.create-on-group-discovery` in `group_watch`, or reframe as `--provider.auto-warm-externally-started` mirroring `auto-stop-externally-started` (reusing its event + reconcile detection)?
2. Coverage: mirroring the auto-stop detection (its periodic scan over all running managed instances) would also catch an instance that ends up running without a session but is already in the group map (no "added" event). Do you see value in that, or do the current add/startup triggers already cover your cases?

Happy to help whichever way is easiest: review, commits on your branch, or co-authoring. Whatever @acouvreur prefers for the final shape, we're glad to do the work.

louisdtt:
Hi, 
I'm pretty sure #985 would work fine for me, i don't mind closing this PR in favour of @antoinemichea solution as it is more customizable from what i see, I can help if needed.

antoinemichea:
Thanks a lot @louisdtt, we really appreciate the collaborative spirit. We'll take the implementation from here, reusing your `seedGroupSession` approach (the `sessions.Get` skip and the `Ready` check are exactly right), and we'll credit you as co-author on the PR. We should be able to manage on our own, but thank you for the offer, and we'll definitely reach out if we end up needing a hand.

No rush to close this PR on your side: happy to leave it open until ours is up so there's no gap, then it can be closed in favor of the converged one. Thanks again!

acouvreur:
Superseded by #993

PR #987, opened 2026-07-01
feat(metrics): expose per-session expiry as a Prometheus gauge

### LINKED ISSUE #986: feat: expose per-session expiry as a Prometheus metric (observability of time-to-shutdown)
**Summary**

Sablier knows exactly when each session expires (`expiresAt` per instance, already serialized by
`tinykv` when `--storage.file` is used), but there is **no live, non-destructive way to observe it**:

- the HTTP API doesn't return expiry, and calling `/api/strategies/*` **renews** the session;
- `--storage.file` is only written on graceful shutdown (no periodic flush) → stale on disk;
- the Prometheus metrics (`--server.metrics.enabled`) are activity counters/histograms only
  (`sablier_session_requests_total`, `sablier_instance_stops_total`, `sablier_instance_*_duration_seconds`, …) — none expose per-session expiry.

This proposes exposing **per-session expiry as a Prometheus metric**, so operators can build a live
"time before shutdown" view.

**Why this matters (use case)**

We run scale-to-zero for a fleet of ~80 on-demand Opencell environments. A recurring operational
question is simply *"how long until env X hibernates?"* — for dashboards, capacity, and debugging.
Today we can only **approximate** it externally (from ingress-router `last-change` timestamps +
the middleware `sessionDuration`), which is noisy: it ignores per-instance session durations and
non-app traffic. Sablier already holds the exact answer; exposing it as a metric would make an
accurate Grafana panel trivial (`expires_at - time()` per group).

**What we built today (approximation) — to show the target UX**

To illustrate the intended dashboard, we already built an *approximate* Grafana table on our side
(derived from Traefik router last-change timestamps + a hardcoded 30m base). It looks like this:

| Environnement | Reste estimé (base 30m) | Inactif depuis |
| --- | --- | --- |
| opencell-x1 | 12.9 min | 17.1 min |
| opencell-x2 | 23.9 min | 6.09 min |
| opencell-x3 | 27.9 min | 2.09 min |
| opencell-x4 | 27.9 min | 2.09 min |
| opencell-x5 | 29.9 min | 5.19 s |

It's genuinely useful, but **noisy by construction**: it assumes a single 30m `sessionDuration` for
every env (long-session envs show wrong/negative values) and only accounts for app-router traffic
(not e.g. keycloak). A native per-session expiry metric would make this **exact and trivial**, with no
guessing.

**Proposed change**

A Prometheus collector that, on each scrape, enumerates the sessions store and emits (non-destructive
read — no `Put`, so it never renews a session):

- `sablier_session_expires_at_timestamp_seconds{instance, group}` — gauge, unix expiry timestamp.
- *(optional)* `sablier_session_seconds_remaining{instance, group}` — convenience; derivable from the
  above in the dashboard, happy to drop it.

Properties:
- gated behind the existing `--server.metrics.enabled`;
- cardinality = number of active sessions (bounded, tens in our case);
- scrape-fresh, and safe (a collector reads state, it doesn't renew).

**Implementation sketch**

Mirrors the existing `GroupLockCollector` (`pkg/metrics/collector.go`), which already reads live
state in `Collect()`. The only missing primitive is a way to enumerate the store: `sablier.Store`
currently exposes `Get/Put/Delete/OnExpire` only. A small read-only enumeration method
(e.g. `Range(func(name string, info InstanceInfo, expiresAt time.Time))` or an `All()` snapshot)
backed by `tinykv` (which already tracks `expiresAt`) would be enough. Then a new
`SessionExpiryCollector` registered alongside `GroupLockCollector`.

Estimate: ~60–120 LOC including tests, no interface change beyond the store enumeration method,
no new dependency.

**Alternative / complementary**

A read-only `GET /api/sessions` endpoint returning `{name, group, expiresAt}` (non-destructive)
would serve CLI/debugging and could feed a custom exporter. Not mutually exclusive with the metric;
the metric is the most Grafana-native path, so that's our preference — but happy to align.

**Additional context**

This is a companion to #985 (grace period for `auto-stop-externally-started`) and composes with
`--storage.file`. As with the CNPG provider work (#943), if this direction works for you we'd be glad
to implement it and open the PR — but only **after** you've validated the design (metric name(s),
labels, store enumeration API). Thanks!

antoinemichea:
## What

Expose the live session expiry as a Prometheus gauge so dashboards can show
"time remaining before scale-to-zero" per instance:

```
sablier_session_expires_at_timestamp_seconds{instance="<name>",group="<group>"} <unix-seconds>
```

One series per **active** session. Emitted by a **non-destructive collector**
that only *reads* the session store at scrape time — it never `Put`s, so
scraping `/metrics` does **not** renew any session. Only active behind the
existing `--server.metrics.enabled` flag; the default behaviour is unchanged.

Closes #986.

## Why

Until now the session expiry was not exposed live anywhere: the HTTP API does
not return it (and reading a session renews it), the `--storage.file` is only
written at shutdown, and the existing metrics are activity counters only. This
gauge makes "remaining time per instance" observable in Grafana.

## Design

- **Store enumeration (non-destructive).** New `Store.Range(ctx, func(InstanceInfo, time.Time)) error`
  that iterates non-expired sessions read-only.
  - `inmemory` delegates to a new `tinykv` `Range` that walks the map under the
    existing lock and skips expired / timeout-less entries (never touches TTLs).
  - `valkey` implements it with `SCAN` + `PTTL` + `GET` (all read-only; TTLs are
    never reset). `expiresAt = now + PTTL`.
- **Collector.** `SessionExpiryCollector` follows the existing `GroupLockCollector`
  pattern (`prometheus.NewDesc` + `MustNewConstMetric` at `Collect` time). To avoid
  an import cycle (`sablier` imports `metrics`) it consumes a small `sessionSource`
  interface; `*Sablier.SessionsSnapshot()` bridges the store to it. The store lock
  is only held to materialise a snapshot, never during metric emission.
- **Registration.** Registered next to `GroupLockCollector` in `start.go`, so it is
  active **only** when `--server.metrics.enabled` is set.
- **`group` label** = `InstanceInfo.Groups[0]` (or `""` when the instance has no group).

### Note from the maintainer, honoured

> "as long as you know that the ExpiresAt is based on the latest access, so it
> would be adjusted every time a session timer is reset."

Yes — this is the intended behaviour. The gauge reflects the current absolute
expiry, so it is pushed back on every renewal. This is documented in the metric
`HELP` string and in `docs/configuration.md`.

### No separate "remaining" metric (intentional)

Remaining time is trivially derivable in Grafana/PromQL and does not need a
second series:

```promql
sablier_session_expires_at_timestamp_seconds - time()
```

Kept the PR minimal.

## Tests

- **Collector unit test** (provider-agnostic): `testutil.CollectAndCompare`
  (exact exposed text, name/labels/value) + `testutil.CollectAndLint` (0 lint
  problems).
- **Non-destructive unit test**: seeds an in-memory store, scrapes the collector
  repeatedly, asserts the exposed `expiresAt` stays constant (a renewing
  collector would make it advance).
- **Store enumeration tests** for `tinykv`, `inmemory` and `valkey` (present vs
  expired entries; `Range` does not renew).
- **Non-regression**: `go build ./...`, `go vet ./...`, `gofmt` (0 diff),
  `golangci-lint run` (0 issues), `go test ./...` (incl. testcontainers).

## End-to-end validation (Podman)

Provider `podman`, `--server.metrics.enabled`, `--sessions.default-duration=2m`.

Before any session — no series (correct):
```
# (no sablier_session_expires_at_timestamp_seconds series)
```

After warming a session (`now ≈ 1782921114`, so expected `≈ 1782921234`):
```
# HELP sablier_session_expires_at_timestamp_seconds Unix timestamp (seconds) at which the instance's session expires and the instance is stopped. ...
# TYPE sablier_session_expires_at_timestamp_seconds gauge
sablier_session_expires_at_timestamp_seconds{group="demo",instance="whoami"} 1.782921234e+09
```

Non-destructive — scraping `/metrics` 4× over 16s of wall-clock leaves the value
unchanged (a renewing collector would advance it):
```
scrape #1 (t=0s):  1.782921234e+09   at 1782921131
scrape #2 (t=6s):  1.782921234e+09   at 1782921137
scrape #3 (t=12s): 1.782921234e+09   at 1782921143
scrape #4 (t=16s): 1.782921234e+09   at 1782921147
```

On expiry (no further warm) the series disappears and the container is stopped:
```
t=1782921231  session_series=1  whoami_status=running
t=1782921241  session_series=0  whoami_status=exited
```

## Notes

- `CHANGELOG.md` is managed by release-please, so it is intentionally not edited
  here — the entry will be generated from the conventional commits.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -286,6 +286,7 @@ When set to `true`, Sablier registers a `GET <base-path>/metrics` route on the s
 |------|------|--------|-------------|
 | `sablier_group_locked` | gauge | `group` | `1` if any instance in the group has an active session, else `0`. One series per known group, including groups with no active sessions. |
 | `sablier_group_active_instances` | gauge | `group` | Number of instances in the group that currently have an active session. |
+| `sablier_session_expires_at_timestamp_seconds` | gauge | `instance`, `group` | Unix timestamp (seconds) at which the instance's session expires and the instance is stopped. One series per active session. The value tracks the latest access, so it is pushed back on every session renewal. Derive the remaining time in Grafana with `sablier_session_expires_at_timestamp_seconds - time()`. |
 | `sablier_instance_start_duration_seconds` | histogram | `instance` | Duration of `provider.InstanceStart` calls (seconds). Observed only on success. |
 | `sablier_instance_ready_duration_seconds` | histogram | `instance` | End-to-end wall time from first not-ready observation to ready (seconds). |
 | `sablier_session_requests_total` | counter | `strategy` (`dynamic`\|`blocking`), `target` (`names`\|`group`) | Total number of session requests received. |
--- a/pkg/metrics/collector.go
+++ b/pkg/metrics/collector.go
@@ -1,6 +1,10 @@
 package metrics
 
-import "github.com/prometheus/client_golang/prometheus"
+import (
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
 
 // GroupsProvider exposes the current configured groups (group name -> instance names).
 type GroupsProvider interface {
@@ -62,5 +66,57 @@ func (c *GroupLockCollector) Collect(ch chan<- prometheus.Metric) {
 		}
 		ch <- prometheus.MustNewConstMetric(c.lockedDesc, prometheus.GaugeValue, locked, group)
 		ch <- prometheus.MustNewConstMetric(c.countDesc, prometheus.GaugeValue, float64(count), group)
+	}
+}
+
+// SessionEntry is the minimal view of a single active session that the
+// SessionExpiryCollector needs to emit its gauge.
+type SessionEntry struct {
+	Instance  string
+	Group     string
+	ExpiresAt time.Time
+}
+
+// sessionSource enumerates the currently active sessions. Implementations must
+// be non-destructive: producing the snapshot must never renew a session.
+type sessionSource interface {
+	SessionsSnapshot() []SessionEntry
+}
+
+// SessionExpiryCollector emits one gauge per active session carrying the unix
+// timestamp (seconds) at which the session expires. It reads the session source
+// lazily at scrape time and never writes back, so scraping /metrics does not
+// renew any session.
+type SessionExpiryCollector struct {
+	sessions sessionSource
+
+	expiresAtDesc *prometheus.Desc
+}
+
+// NewSessionExpiryCollector wires a sessionSource into a Collector that can be
+// registered on the same registry as the rest of the metrics.
+func NewSessionExpiryCollector(sessions sessionSource) *SessionExpiryCollector {
+	return &SessionExpiryCollector{
+		sessions: sessions,
+		expiresAtDesc: prometheus.NewDesc(
+			"sablier_session_expires_at_timestamp_seconds",
+			"Unix timestamp (seconds) at which the instance's session expires and the instance is stopped. One series per active session. The value tracks the latest access and is pushed back on every session renewal. Derive the remaining time in Grafana with the expression: value - time().",
+			[]string{"instance", "group"}, nil,
+		),
+	}
+}
+
+func (c *SessionExpiryCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- c.expiresAtDesc
+}
+
+func (c *SessionExpiryCollector) Collect(ch chan<- prometheus.Metric) {
+	for _, s := range c.sessions.SessionsSnapshot() {
+		ch <- prometheus.MustNewConstMetric(
+			c.expiresAtDesc,
+			prometheus.GaugeValue,
+			float64(s.ExpiresAt.Unix()),
+			s.Instance, s.Group,
+		)
 	}
 }
--- a/pkg/metrics/collector.go
+++ b/pkg/metrics/collector.go
@@ -77,9 +77,9 @@ type SessionEntry struct {
 	ExpiresAt time.Time
 }
 
-// sessionSource enumerates the currently active sessions. Implementations must
+// SessionSource enumerates the currently active sessions. Implementations must
 // be non-destructive: producing the snapshot must never renew a session.
-type sessionSource interface {
+type SessionSource interface {
 	SessionsSnapshot() []SessionEntry
 }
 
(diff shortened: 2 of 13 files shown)


acouvreur:
Thanks for your pull request @antoinemichea, I think it makes a lot of sense.

Did you have the chance to compile the code yourself and use it in your infrastructure ? I would be very curious to know if you currently have a dashboard with the gauges you've implemented.

acouvreur:
@jindrichskupa do you think you can take a look at this pull request ? I know you've worked a lot on the metrics for Sablier. Would really appreciate your review here.

antoinemichea:
Hi @acouvreur , yes i deployed local version on my cluster here is a sample of my grafana dashbord : image

acouvreur (changes requested):
I think InstanceInfo should contain it's expiration date, but I'm not sure how complicated it would be to actually do that.

In the meantime, simply make Range create a instant snapshot, and run the callback on each item of the snapshot.

antoinemichea:
Thanks! Pushed the review changes:

- **`Range` now snapshots then iterates** (tinykv/in-memory): it collects the non-expired entries under the store lock, releases the lock, and only then runs the callback per item — so the callback can be slow or re-enter the store, and the store mutex is no longer held during it.
- **valkey `Range`**: the absolute expiry is now captured from `PTTL` immediately, before the `GET` + unmarshal, so a slow `GET` under load can't push the reported expiry later than the real TTL (Copilot's point).
- **Exported `SessionSource`** so external packages can assert conformance; added `var _ metrics.SessionSource = (*sablier.Sablier)(nil)`.

On carrying the expiry inside `InstanceInfo`: agreed it'd be cleaner, but it's a larger change touching every store's write path (the TTL would have to be stamped into the value on `Put`), so I kept it out of this PR — happy to do it as a follow-up if you'd like.

For the deployment question: this is running in production now (a fleet of ~80 on-demand environments), combined with #993. The gauge feeds a Grafana panel showing exact remaining time per environment via `sablier_session_expires_at_timestamp_seconds - time()` (screenshot above).

One more thing surfaced while running it: I found and fixed a pre-existing expiry bug in `tinykv` (unrelated to this PR) that could drop the `onExpire` notification for a session, leaving an instance running with no session — opened separately as #995.

acouvreur (approved):
Tests failures are not related to this feature

Now merging

Issue #975, opened 2026-06-20
Feature Request: Support containers that don't need healthcheck waiting


ngoc-minh-do:
I have some setups where a container depends on a main container, but the supporting containers do not need to be healthy before the main container is considered ready.

Currently, even after disabling the Docker healthcheck, Sablier still shows the error screen:

container exited with code "145"

It would be useful to have a way to configure certain containers as "non-blocking" dependencies, so Sablier only waits for the main container (or selected containers) to become ready.

Thanks!

acouvreur:
I believe this is slightly related to https://github.com/sablierapp/sablier/issues/952

Basically Sablier should honor the restart policies. Right now I still wanted to make it require an exit status 0, but you'd need Sablier to start your container and not care about it basically ?

ngoc-minh-do:
I think my scenario is simpler than #952. I only need Sablier to start the container when the URL is accessed (if it is not already running), without waiting for a healthcheck.

I believe this worked in previous versions by disabling the Docker healthcheck. However, now it briefly (may be 1 second) shows the waiting screen with `container exited with code "145"`

ngoc-minh-do:
@acouvreur By the way, can I ask why the `container exited with code "145"` shown briefly before `starting` message? I think in few previous version, the first message was always  `starting`

acouvreur:
> [@acouvreur](https://github.com/acouvreur) By the way, can I ask why the `container exited with code "145"` shown briefly before `starting` message? I think in few previous version, the first message was always `starting`

Containers failing to start on repeat now show the latest status from the previous invocation, restarting it on access.

Otherwise, if a container would be failing continuously, it would be restarted continuously as well without providing any feedback.

acouvreur:
Completed in #984

PR #999, opened 2026-07-05
feat(api): add ready_on_start query param to dynamic strategy

ngoc-minh-do:
## What

Adds `ready_on_start` query parameter to the dynamic strategy. When `true`, Sablier returns `X-Sablier-Session-Status: ready` immediately regardless of instance health — the reverse proxy passes the request through without showing a waiting page.

The per-instance `sablier.ready-on-start` label controls behavior per container. The query param does the same, but at the middleware level — no labels needed.

## Why

Sometimes you want different behavior depending on which service is accessed:

**Home Assistant + Frigate:** When loading the HA dashboard, you want Frigate to start in the background — no need to wait. But when accessing Frigate's own UI directly, you want the waiting screen to confirm it's ready.
**Chat app + LiteLLM proxy:** When users open the chat, LiteLLM starts alongside without blocking. Direct access to LiteLLM UI get the waiting screen.

## Usage

GET /api/strategies/dynamic?group=myapp&ready_on_start=true

Plugin-side config (separate PRs):
```yaml
    frigate-sablier:
      plugin:
        sablier:
          names: frigate
          dynamic:
            displayName: Frigate

    frigate-sablier-force:
      plugin:
        sablier:
          names: frigate
          dynamic:
            displayName: Frigate
            readyOnStart: true
```

#975 #984

### DIFF
--- a/internal/api/start_dynamic.go
+++ b/internal/api/start_dynamic.go
@@ -23,6 +23,7 @@ type DynamicRequest struct {
 	Theme            string        `form:"theme"`
 	SessionDuration  time.Duration `form:"session_duration"`
 	RefreshFrequency time.Duration `form:"refresh_frequency"`
+	ReadyOnStart     bool          `form:"ready_on_start"`
 }
 
 func StartDynamic(router *gin.RouterGroup, s *ServeStrategy) {
@@ -74,6 +75,9 @@ func StartDynamic(router *gin.RouterGroup, s *ServeStrategy) {
 		}
 
 		AddSablierHeader(c, sessionState)
+		if request.ReadyOnStart {
+			c.Header(SablierStatusHeader, SablierStatusReady)
+		}
 
 		renderOptions := theme.Options{
 			DisplayName:      request.DisplayName,
--- a/internal/api/start_dynamic_test.go
+++ b/internal/api/start_dynamic_test.go
@@ -27,6 +27,22 @@ func session() *sablier.SessionState {
 	}
 }
 
+func notReadySession() *sablier.SessionState {
+	return &sablier.SessionState{
+		Instances: map[string]sablier.InstanceInfoWithError{
+			"test": {
+				Instance: sablier.InstanceInfo{
+					Name:            "test",
+					CurrentReplicas: 0,
+					DesiredReplicas: 1,
+					Status:          sablier.InstanceStatusStarting,
+				},
+				Error: nil,
+			},
+		},
+	}
+}
+
 func TestStartDynamic(t *testing.T) {
 	t.Run("StartDynamicInvalidBind", func(t *testing.T) {
 		app, router, strategy, _ := NewApiTest(t)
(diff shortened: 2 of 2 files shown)


acouvreur:
If you want to change the behavior of the dynamic strategy that much that the waiting page does not exist, maybe this should simply be another strategy ?

I initially created https://github.com/sablierapp/sablier/pull/919 for that purpose. A dedicated "poke" API that returns instantly and wake up / extend the session duration.

Right now the poke API was designed exclusively for extending the session, but starting would be the same behaviour.

Here we're facing three strategies:
- Waiting Page
- Blocking
- None

But can you explain more in detail why you would apply this at the group level ? Because, if we pass-through instantly, then none of the services are ready to receive any request.

ngoc-minh-do:
In the PR, I use Frigate as an example of a standalone container, but it could also represent a group of containers.

For example, if we integrate Immich (which consists of four containers) with Home Assistant:

* When we access Home Assistant, it should serve requests without waiting for Immich.
* However, when we access Immich, we want to wait until all four containers are ready.

Does this make sense?

acouvreur:
> In the PR, I use Frigate as an example of a standalone container, but it could also represent a group of containers.
> 
> For example, if we integrate Immich (which consists of four containers) with Home Assistant:
> 
> * When we access Home Assistant, it should serve requests without waiting for Immich.
> * However, when we access Immich, we want to wait until all four containers are ready.
> 
> Does this make sense?

It does make sense, but in the case of Home Assistant, Home Assistant itself will also be considered ready instantly.

Which will make all subsequent requests to fail because HA is not really ready.

What you want here is mixing readiness statuses on different containers I believe, not the whole group.

What you are doing here, putting the query parameter at the API level, applies it to the whole group.

ngoc-minh-do:
In my imagination, the config would be something like this. I always separate containers group by project/service, so Home assistant and Immich or Frigate will be in different group

```yaml
http:
  routers:
    homeassistant:
      ...
      middlewares:
        - homeassistant-sablier
        - immich-force-sablier

    immich:
      ...
      middlewares:
        - immich-sablier

  middlewares:
    homeassistant-sablier:
      plugin:
        sablier:
          group: homeassistant
          dynamic:
            displayName: homeassistant

    immich-force-sablier:
      plugin:
        sablier:
          group: immich
          dynamic:
            readyOnStart: true

    immich-sablier:
      plugin:
        sablier:
          group: immich
          dynamic:
            displayName: immich

```

acouvreur:
Never thought of chaining middlewares, that's quite clever 😀

But in that case, I think we should definitely leverage a new strategy "none"/"poke": meaning there's no need to create such options on top of existing strategy that does not make sense.

ngoc-minh-do:
I agree. 
I have some service that chains multiple sablier middleware. In the past, it worked ok, but now since each middleware always shows the waiting screen briefly, I have to see multiple waiting screens until it reaches the destination screen. That's why I want this feature so bad :D

ngoc-minh-do:
> Never thought of chaining middlewares, that's quite clever 😀
> 
> But in that case, I think we should definitely leverage a new strategy "none"/"poke": meaning there's no need to create such options on top of existing strategy that does not make sense.

Which new strategy name do you prefer, or should I just choose one?

acouvreur:
> > Never thought of chaining middlewares, that's quite clever 😀
> > 
> > But in that case, I think we should definitely leverage a new strategy "none"/"poke": meaning there's no need to create such options on top of existing strategy that does not make sense.
> 
> Which new strategy name do you prefer, or should I just choose one?

Yes please give some suggestions.

Maybe something like poke or none is okay

ngoc-minh-do:
What about 

* `poke`
* `none`
* `bypass`
* `no-wait`
* `direct`
* `pass-through`
* `async`
* `background`
* `prewarm`
* `fire-and-forget`
* `startup-only`
* `non-blocking`
* `detached`
* `deferred`
* `immediate`
* `lazy-start`
* `warmup`
* `wake`
* `activate`
* `resume`
* `bootstrap`
* `dispatch`
* `trigger`
* `signal`
* `notify`

acouvreur:
> What about
> 
>     * `poke`
> 
>     * `none`
> 
>     * `bypass`
> 
>     * `no-wait`
> 
>     * `direct`
> 
>     * `pass-through`
> 
>     * `async`
> 
>     * `background`
> 
>     * `prewarm`
> 
>     * `fire-and-forget`
> 
>     * `startup-only`
> 
>     * `non-blocking`
> 
>     * `detached`
> 
>     * `deferred`
> 
>     * `immediate`
> 
>     * `lazy-start`
> 
>     * `warmup`
> 
>     * `wake`
> 
>     * `activate`
> 
>     * `resume`
> 
>     * `bootstrap`
> 
>     * `dispatch`
> 
>     * `trigger`
> 
>     * `signal`
> 
>     * `notify`

I feel like a poke strategy would be perfect. That was already the word I was going for in another pull request.

Would you be able to create the new strategy ?

ngoc-minh-do:
Sure! I agree that poke fits the behavior well. I'll implement the new poke strategy

acouvreur:
Closing in favor of a poke strategy

ngoc-minh-do:
poke is on #1015 (just for reference)

PR #1002, opened 2026-07-07
fix(kubernetes): detect CloudNativePG CRD via discovery instead of List

### LINKED ISSUE #1001: Kubernetes: CNPG CRD detection via List turns missing RBAC into a permanent reflector error loop
**Describe the issue**

`clusterCRDInstalled` (pkg/provider/kubernetes/cnpgcluster_events.go) probes for the CloudNativePG CRD by **listing** `postgresql.cnpg.io/v1 clusters`. That requires RBAC on the resource itself. On clusters that do **not** run CloudNativePG and whose sablier ClusterRole is scoped to deployments/statefulsets (least privilege), the probe fails with `Forbidden` - which the guard treats as "assume the CRD exists" and starts the dynamic informer anyway. The reflector then retries forever:

```
E... reflector.go:227 "Failed to watch" err="failed to list postgresql.cnpg.io/v1, Resource=clusters: clusters.postgresql.cnpg.io is forbidden: User \"system:serviceaccount:sablier:sablier\" cannot list resource \"clusters\"..."
```

Granting `get/list/watch` on the (non-existent) resource only changes the loop to `404 the server could not find the requested resource` until the pod is restarted, because the guard is evaluated once at startup.

**Context**
 - Sablier version: 1.14.0
 - Provider: Kubernetes (AKS), CloudNativePG **not** installed
 - Reverse proxy: Traefik plugin

**Expected behavior**

Detecting the CRD should not require RBAC on the CRD''s resources. The discovery API (`ServerResourcesForGroupVersion`) is readable by any authenticated service account, cleanly distinguishes "group/version not served" from transient errors, and avoids the Forbidden ambiguity entirely.

PR with that change: incoming.

thisispr:
Fixes #1001

## Problem

`clusterCRDInstalled` probes for the CNPG CRD by listing `postgresql.cnpg.io/v1 clusters`, which requires RBAC on the resource. With a least-privilege ClusterRole (deployments/statefulsets only) on a cluster **without** CloudNativePG, the probe returns `Forbidden`, the guard assumes the CRD exists, and the dynamic informer runs a permanent list/watch error loop (`reflector.go "Failed to watch" ... is forbidden` / after granting RBAC: `404 the server could not find the requested resource` until restart, since the guard runs once at startup).

## Change

Probe via the **discovery API** (`Client.Discovery().ServerResourcesForGroupVersion`) and check the resource is served:

- needs no RBAC on the CRD''s resources (discovery is readable by any authenticated SA), so least-privilege deployments/statefulsets-only roles work out of the box
- `NotFound` (group/version not served) -> watcher skipped, matching the intent of the existing guard
- other errors keep the previous fail-open behavior (warn + enable watcher)

`go build` / `go vet` clean; package tests need a Docker/k3s testcontainer which my environment lacks - relying on CI. Observed/verified the failure mode on an AKS cluster running sablier 1.14.0 with Traefik.


acouvreur:
@antoinemichea does this look good to you ?

antoinemichea:
@acouvreur yes, LGTM this is the right fix, and it's the correct fix for the guard that gates the CNPG hibernation watcher we added in #944.

Reviewed the change against the k8s discovery semantics:

- **Discovery is the correct probe.** `ServerResourcesForGroupVersion` needs no RBAC on `clusters.postgresql.cnpg.io` (discovery is served to any authenticated SA via the default `system:discovery` binding), so a least-privilege `deployments/statefulsets` ClusterRole works out of the box — which is exactly the #1001 scenario. A missing group/version returns a 404 that reliably satisfies `apierrors.IsNotFound`, so the `IsNotFound → false` branch is contractually sound.
- **Targeting a single GroupVersion is the right call** (vs full discovery): it does one `GET /apis/postgresql.cnpg.io/v1` and can't hit the `ErrGroupDiscoveryFailed` partial-failure trap when some unrelated aggregated APIService is down.
- **No false positives/negatives**: the exact `r.Name == "clusters"` match won't be fooled by subresources (`clusters/status`), the probed GVR matches the informer's GVR, and the `p.Client == nil` guard is correctly added now that `Discovery()` is dereferenced. The non-`NotFound` fail-open is preserved, so a cluster that *does* run CNPG but is missing list/watch RBAC still enables the informer and surfaces the error as before — no behavior change there.

Two non-blocking notes:

1. A small unit test would lock this in without needing a testcontainer, a `fake.NewSimpleClientset()` with `Fake.Resources` set lets you assert all three branches (served → true, `NotFound` → false, other error → true) plus the nil guards. Nice-to-have, not a blocker.
2. Minor: `ServerResourcesForGroupVersion` doesn't take a `ctx` (uses `context.TODO()` internally), so the `ctx` passed to `clusterCRDInstalled` now only feeds the log line, not the HTTP GET. It's a one-shot startup call bounded by the REST client timeout, so impact is negligible, just noting it.

👍 from me.

acouvreur (dismissed):
LGTM Thanks!

acouvreur:
Please address lint issues and then will merge :)

acouvreur (approved):
LGTM Thanks!

Issue #258, opened 2024-02-14
Always scales back up to 1 replica on Kubernetes


dimber-cais:
**Describe the issue**
When Sablier scales up a previously scaled down deployment, it always scales it to 1 pod, and does not return the deployment back to the value of `replicas` specified in the original YAML.

**Context**
 - Sablier version: All
 - Provider: Kubernetes
 - Reverse proxy: N/A
 - Sablier running inside a container? Yes

**Expected behavior**
It should be possible to scale back up to multiple replicas when scaling back up from zero. This could either be by capturing the original `spec.replicas` for the deployment and using that, or introducing a new `sablier.replicas` label.

**Additional context**
This would allow Sablier to be used in scenarios where workloads should both scale to zero when not in use and scale to a highly available multi-instance deployment when in use. 

I could try to contribute this as a PR if you are open for this......

dimber-cais:
I think [this](https://github.com/acouvreur/sablier/blob/main/app/providers/kubernetes.go#L123-L124) is the applicable code.

acouvreur:
Hello @dimber-cais,

Do you know how I can retrieve the original replicas ?

What happens if the original replicas is 0 ?

I can add this fairly easily with groups, however the API using `names` query param would not support it properly, or it would scale all mentioned workloads to this specific replicas.

PR #1003, opened 2026-07-07
fix(kubernetes): honor names-based replica count on instance start

thisispr:
## Problem

The names-based API documents `kind_namespace_name_replicas` (README, e2e manifests use e.g. `deployment_default_whoami-deployment_1`), and `ParseName` extracts the trailing count - but `InstanceStart` discards it: the target always comes from `ScaleConfigFromLabels`, which defaults to `Active.Replicas = 1` when the workload has no `sablier.active.replicas` label.

Anyone configuring `deployment_ns_name_2` via a reverse-proxy plugin middleware gets 1 replica on wake, silently. This is the still-open half of #258: that issue was closed by the scale-mode labels (#908), which works for label-managed workloads, but the names path - the primary interface for plugin middleware users, who may not control workload labels (third-party charts) - still ignores its own documented replicas segment. Observed on sablier 1.14.0 / Traefik plugin / AKS: `deployment_<ns>_<name>_2` woke at 1, scaling a running 2-replica deployment down.

## Change

`InstanceStart` resolves the target count via a small helper with explicit precedence:

1. `sablier.active.replicas` label, when present - **stays authoritative** (no behavior change for label users; an explicit label of `1` still wins over a names count)
2. an explicit names count `> 1`
3. the existing default (1)

Includes a table-driven unit test for the precedence (`TestActiveReplicas`). `go build`/`go vet` clean; the package suite needs the k3s testcontainer my environment cannot run - relying on CI.

## Notes

- `_0` and `_1` behave exactly as before (default), so existing setups are unaffected.
- The DeploymentName/StatefulSetName helpers (label-discovered workloads) keep their current behavior; only explicitly-named instances change, and only when the name asks for > 1.

### DIFF
--- a/pkg/provider/kubernetes/instance_start.go
+++ b/pkg/provider/kubernetes/instance_start.go
@@ -16,6 +16,22 @@ func (p *Provider) InstanceDependencies(_ context.Context, _ string) ([]sablier.
 	return nil, nil
 }
 
+// activeReplicas resolves the replica count used when starting an instance.
+// The sablier.active.replicas label on the workload stays authoritative when
+// present. Without it, an explicit count from the names-based API
+// (kind_namespace_name_replicas, e.g. reverse-proxy plugin middlewares) is
+// honored, so `deployment_ns_name_2` scales to 2 as documented instead of
+// silently falling back to the default of 1.
+func activeReplicas(sc sablier.ScaleConfig, labels map[string]string, parsed ParsedName) int32 {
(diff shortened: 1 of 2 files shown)


acouvreur (approved):
LGTM

acouvreur:
Thank you! Even though I want to drop support for `names` in the API, I'm ok with that fix.

acouvreur (approved):

PR #1015, opened 2026-07-10
feat(strategy): add poke strategy

ngoc-minh-do:
## What

Adds the **poke strategy** — a new `GET /api/strategies/poke` endpoint that starts instances and returns the session status immediately without waiting for readiness. The plugin always passes the request through regardless of `X-Sablier-Session-Status`.

Useful when you want instances started but don't need them healthy before serving traffic — background workers, sidecar caches, companion processes the caller doesn't depend on.

## Usage

GET /api/strategies/poke?group=myapp&session_duration=30m

Plugin config examples (separate PRs in each plugin repo):

```yaml
# Traefik
plugin:
  sablier:
    group: myapp
    poke: {}
# Caddy
sablier {
    group mygroup
    poke
}
// Proxy-WASM
{ "group": "mygroup", "poke": {} }
```

Related
- sablier.ready-on-start=true label (already merged) — same behavior at the per-container level
- Companion PRs in plugin repos: sablier-traefik-plugin, sablier-caddy-plugin, sablier-proxywasm-plugin

### DIFF
--- a/docs/content/concepts/strategies.md
+++ b/docs/content/concepts/strategies.md
@@ -52,6 +52,8 @@ This strategy is useful for background services whose startup the caller does no
 
 Instances with the `sablier.ready-on-start=true` label are treated as ready as soon as the start is dispatched, even with the dynamic or blocking strategy. See [Configuration](/configuration/#instance-labels) for details.
 
+To set up the poke strategy on a route, see [Start without waiting](/how-to-guides/loading-strategies/start-without-waiting/).
+
 ```mermaid
 sequenceDiagram
     participant User
--- a/docs/content/concepts/strategies.md
+++ b/docs/content/concepts/strategies.md
@@ -3,7 +3,7 @@ title: Strategies
 weight: 418
 ---
 
-A **strategy** is how the reverse-proxy plugin behaves while Sablier is starting the instances for a session. Sablier offers two: the **dynamic** strategy shows a self-refreshing waiting page, and the **blocking** strategy holds the request open until everything is ready. Which one fits depends on who (or what) is making the request.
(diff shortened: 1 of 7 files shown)


acouvreur:
>  and immediately returns `X-Sablier-Session-Status: ready`, skipping the health check.

I'm not sure we should do that. It should be up to the consumer to understand what the poke api does. We should not "lie" to the consumer telling them this is ready.

~~I think we should even return `X-Sablier-Session-Status: unknown`~~

We should return the actual status of the session. Then, the poke usage in the proxy could allow: `continue on error` for example.

ngoc-minh-do:
> I'm not sure we should do that. It should be up to the consumer to understand what the poke api does. We should not "lie" to the consumer telling them this is ready.
> 
> ~I think we should even return `X-Sablier-Session-Status: unknown`~
> 
> We should return the actual status of the session. Then, the poke usage in the proxy could allow: `continue on error` for example.

Good point. So i will make Sablier return actual status, plugin side will bypasses of strategy is poke

acouvreur:
Can you add the openapi description in comments like it was done in the other routes ? Thanks, and then don't forget to run the generate command so that the documentation gets build automatically for you

ngoc-minh-do:
> Can you add the openapi description in comments like it was done in the other routes ? Thanks, and then don't forget to run the generate command so that the documentation gets build automatically for you

done

acouvreur:
@ngoc-minh-do let's rebase this PR and merge it

ngoc-minh-do:
@acouvreur I rebased it, could you approve the workflows

acouvreur:
@ngoc-minh-do can you fix the doc mismatch ? That'd be perfect

ngoc-minh-do:
@acouvreur Fixed.

acouvreur (approved):

PR #1023, opened 2026-07-11
refactor(sablier): stage 1 of the InstanceInfo split - typed InstanceConfig with one parser and one enabled-semantics

acouvreur:
## Finding

`InstanceInfo` ([instance.go](../blob/main/pkg/sablier/instance.go#L40-L86)) is **four things at once**:

1. **Provider-reported state** - `Status`, `CurrentReplicas`, `DesiredReplicas` (plus `ReadyAt`, whose own comment admits it "is set internally by Sablier and is never populated by a provider").
2. **Parsed-label configuration bag** - `Enabled` is a *string* compared against `"true"` in **six** scattered places (autostop, autowarm, expiration, request handling, running-hours watcher, and the parser itself); `RunningHours`/`RunningDays` are raw strings; `ReadyAfter`, `ReadyOnStart`, `AntiAffinity`, `ScaleConfig` ride along.
3. **The API wire format** - marshalled directly into responses and the SSE stream; every field is OpenAPI schema surface that reverse-proxy plugins can parse.
4. **The persisted session record** - the whole struct is JSON-serialized into the store, freezing stale label config into sessions.

Every feature adds a field here and touches all four concerns at once; nobody can tell which fields a provider must set, which Sablier owns, and which are configuration. Label parsing also warned through the **global** `slog`, bypassing the configured handler.

## Approach: staged, wire-compatible

The full split has two compatibility cliffs - the wire format (plugins) and the persisted records (upgrades with live state) - so it lands in stages. **This PR is stage 1, scoped to the configuration concern, and changes zero existing bytes on the wire.**

## What this stage does

* **One parser.** All `sablier.*` label interpretation now lives in `InstanceConfigFromLabels`, producing a typed `InstanceConfig` (`Enabled bool`, real `time.Duration`, validated window strings, `Scale *ScaleConfig`). It takes a logger (restoring per-instance context on warnings) instead of the global `slog`.
* **One semantics.** The six `Enabled == "true"` string comparisons collapse into `IsEnabled()` accessors on `InstanceInfo` and `InstanceConfiguration`. The `InstanceInfo` accessor prefers the parsed config and **falls back to the flat field** for session records persisted by older versions.
* **Compatibility preserved.** `PopulateEnabledAndGroup` (the function all five providers call) becomes a thin adapter: it stores the typed config under a new **additive** `config` key and mirrors it into the flat legacy fields byte-identically - `enabled` stays the raw label string the API has always exposed.

## Evidence it is behavior-preserving

* Every pre-existing `PopulateEnabledAndGroup` test (ready-after, groups, running-hours, anti-affinity, scale labels, blkio labels) **passes unchanged** - they assert the flat fields.
* A new `TestPopulateEnabledAndGroup_WireCompat` pins the contract: legacy keys keep their exact names and raw values, `config` is strictly additive.
* A new fallback test unmarshals a pre-`config` session record and asserts `IsEnabled()` still answers correctly.
* Full parser table test for `InstanceConfigFromLabels` (enable exact-match semantics, group gating, invalid-value warn-and-ignore, scale-presence condition).

`openapi.json` regenerated (gains the additive `sablier.InstanceConfig` schema); `make check-generate`, full test suite and `golangci-lint` pass.

## Next stages (separate PRs)

2. **API DTO in `internal/api`** so the wire format stops being the domain struct (pairs with the OpenAPI drift finding).
3. **Versioned session record** so the store stops persisting label configuration, and the flat config fields can then be dropped from `InstanceInfo`.
4. Provider contract cleanup (`ReadyAt` and friends move off the provider-facing type).

## Docs

No documentation impact beyond the regenerated OpenAPI spec: no labels, config options, or behavior changed.

### DIFF
--- a/docs/static/openapi.json
+++ b/docs/static/openapi.json
@@ -80,6 +80,57 @@
         },
         "type": "object"
       },
+      "sablier.InstanceConfig": {
+        "properties": {
+          "antiAffinity": {
+            "description": "AntiAffinity lists the groups this instance backs off from\n(sablier.anti-affinity).",
+            "items": {
+              "type": "string"
+            },
+            "type": "array"
+          },
+          "enabled": {
+            "description": "Enabled reports whether the instance opted into Sablier management\n(sablier.enable set to exactly \"true\").",
+            "type": "boolean"
+          },
+          "groups": {
+            "description": "Groups the instance belongs to (sablier.group). Only populated for\nenabled instances; defaults to [\"default\"] when the label is absent.",
+            "items": {
+              "type": "string"
+            },
+            "type": "array"
+          },
+          "readyAfter": {
+            "allOf": [
+              {
+                "$ref": "#/components/schemas/time.Duration"
+              }
+            ],
+            "description": "ReadyAfter is the settling delay after the instance first reports ready\n(sablier.ready-after). Zero means no extra wait."
+          },
+          "readyOnStart": {
+            "description": "ReadyOnStart marks the instance ready as soon as its start is\ndispatched, skipping the health check (sablier.ready-on-start).",
+            "type": "boolean"
+          },
+          "runningDays": {
+            "description": "RunningDays restricts RunningHours to specific weekdays\n(sablier.running-days). Empty means every day.",
+            "type": "string"
+          },
+          "runningHours": {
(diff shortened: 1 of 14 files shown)


acouvreur:
CI failures fixed: the provider inspect suites DeepEqual the full InstanceInfo and needed the new additive `Config` field in their expectations. Since `PopulateEnabledAndGroup` mirrors the parsed config 1:1 into the flat fields each case already declares, the expectation is now derived from those fields at the per-case fixup block (works for the K8s annotation cases too). Verified locally against dind (docker + swarm), k3s and pind - all four inspect suites pass.

PR #1041, opened 2026-07-24
fix(dockerswarm): honor the running-only filter in InstanceList

bondarslavik:
## Problem

The Docker Swarm provider ignores `InstanceListOptions` and returns every `sablier.enable=true` replicated service, including services already scaled to zero:

```go
func (p *Provider) InstanceList(ctx context.Context, _ provider.InstanceListOptions) (...)
```

`StopAllUnregisteredInstances` explicitly asks for running instances only (`All: false`). With `provider.auto-stop-externally-started: true` the 30s reconciliation scan therefore treats every sleeping (scaled-to-zero, sessionless) service as an externally started instance and "stops" it again on every tick. For each sleeping environment this produces, every 30 seconds, forever:

- one `INF stopped unregistered instance ... reason="instance is enabled but not started by Sablier"` log line per service (~26k lines/day for a 9-service stack), and
- one no-op `ServiceUpdate` per service, bumping the service spec version and generating docker events churn.

Sample from a real deployment (every 30s, all services already at 0/0):

```
1:41PM INF sablier/autostop.go:53 stopped unregistered instance instance=feature-x_queue-all reason="instance is enabled but not started by Sablier"
1:41PM INF sablier/autostop.go:53 stopped unregistered instance instance=feature-x_queue    reason="instance is enabled but not started by Sablier"
... (9 lines per stack, repeated every scan)
```

## Fix

Mirror the docker provider semantics (`ContainerList` with `All: options.All`): when `All` is `false`, skip services that have no running tasks. `ServiceStatus` is already populated since the listing passes `Status: true`.

## Testing

- Added `TestDockerSwarmProvider_InstanceList_OnlyRunning`: a running service and a scaled-to-zero service, `All: true` returns both, `All: false` returns only the running one (needed a `Replicas` knob on the `MimicOptions` test helper, defaulting to the previous hardcoded 1).
- `go test ./pkg/provider/dockerswarm/` (full testcontainers suite) and `go test ./pkg/sablier/ -short` pass.

A possible follow-up (not in this PR to keep it focused): `ServiceUpdateScale` could skip the `ServiceUpdate` call entirely when the current spec already matches the requested replicas/resources, making stops idempotent at the API level too.

### DIFF
--- a/pkg/provider/dockerswarm/service_list.go
+++ b/pkg/provider/dockerswarm/service_list.go
@@ -11,12 +11,12 @@ import (
 	"github.com/sablierapp/sablier/pkg/sablier"
 )
 
-func (p *Provider) InstanceList(ctx context.Context, _ provider.InstanceListOptions) ([]sablier.InstanceConfiguration, error) {
+func (p *Provider) InstanceList(ctx context.Context, opts provider.InstanceListOptions) ([]sablier.InstanceConfiguration, error) {
 	filters := client.Filters{}
 	filters.Add("label", fmt.Sprintf("%s=true", sablier.LabelEnable))
 	filters.Add("mode", "replicated")
 
-	p.l.DebugContext(ctx, "listing services", slog.Group("options", slog.Bool("status", true), slog.Any("filters", filters)))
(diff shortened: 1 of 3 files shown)


acouvreur (approved):
LGTM!

Thank you very much for your fix and investigation!

acouvreur (approved):

PR #1068, opened 2026-08-11
feat(session): best-effort instances via the `optional:` name prefix

thisispr:
## Description

Prefixing a requested instance name with `optional:` marks it as **best-effort**: the instance is started and kept alive with the session exactly like any other member, but its readiness and errors never gate the session. The waiting page clears (and a blocking request returns) as soon as all non-optional instances are ready — even if an optional instance is still starting, unhealthy, or failed to start entirely.

```yaml
names: whoami,optional:whoami-helper
```

## Motivation

We run Sablier (Kubernetes provider, Traefik plugin, `names`-based middlewares) to scale entire dev environments (~80 deployments each) to zero. A single crashlooping auxiliary deployment (reporting worker, devtools API, …) currently holds the whole environment on the waiting page forever. `sablier.ready-on-start` helps for slow health checks but requires labeling every workload's chart, doesn't cover third-party charts, and still gates on start errors.

This feature lets a `names`-based setup split a stack into important instances visitors must wait for and auxiliary ones that wake and hibernate together with the stack but can never hold it hostage when broken.

## Design notes

- **The marker travels inside the name**, so every existing reverse-proxy plugin forwards it unchanged — no plugin change, no plugin upgrade needed by users.
- **Collision-free**: Docker, Swarm, Podman and Kubernetes names cannot contain a colon.
- **Stripped early**: the prefix is removed at session-request time (`RequestSession`), so providers, the session store, metrics and the waiting page all see the clean name.
- Optionality is **per request**, carried in `InstanceInfoWithError.Optional`; `SessionState.IsReady()`, `InstanceErrors()` and `NotReadyInstances()` skip optional entries. A session whose members are all optional is immediately ready.
- Optional instances still appear in session/API responses (new `optional` field on the instance entry) so the waiting page details remain complete.
- Group-based sessions are unchanged; a `sablier.best-effort` label for label-discovered groups would be the natural follow-up if this direction is accepted — happy to do that in a separate PR.

## Changes

- `pkg/sablier/session_request.go` — `OptionalPrefix` constant, prefix stripping, `Optional` flag on `InstanceInfoWithError`
- `pkg/sablier/session.go` — optional entries skipped in `IsReady`, `NotReadyInstances`, `InstanceErrors`
- `internal/api` — `optional` field on `InstanceEntryResponse`, `names` param docs for both strategies
- `docs/content/how-to-guides/readiness/optional-instances.md` — new how-to guide
- `docs/static/openapi.json` — regenerated (`make openapi`)

## Tests

- `TestSessionState_OptionalInstancesDoNotGate`, `TestSessionState_RequiredInstanceStillGatesAlongsideOptional`, `TestSessionState_InstanceErrors_SkipsOptional`
- `TestRequestSession_OptionalPrefix` — verifies the prefix is stripped before store/provider calls, the entry is keyed by the clean name, and carries the flag

`go build ./...`, `go vet` and the `pkg/sablier` + `internal/api` suites pass locally.

### DIFF
--- a/docs/content/how-to-guides/readiness/optional-instances.md
+++ b/docs/content/how-to-guides/readiness/optional-instances.md
@@ -1,0 +1,54 @@
+---
+title: Mark instances as optional (best-effort)
+description: Start an instance with the session without letting its readiness or failures block the session.
+weight: 137
+compatibility:
+  docker: supported
+  swarm: supported
+  kubernetes: supported
+  podman: supported
+  proxmox: supported
+---
+
+{{< compatibility >}}
+
+This guide shows you how to mark an instance as **optional** (best-effort) by prefixing its name with `optional:` in a `names` request:
+
+```yaml
+# Traefik middleware example
+http:
+  middlewares:
+    my-sablier:
+      plugin:
+        sablier:
+          sablierUrl: http://sablier:10000
+          names: whoami,optional:whoami-helper
+          dynamic:
+            displayName: My Stack
+```
+
+An optional instance behaves exactly like any other session member — it is started on demand, its session is refreshed by traffic, and it scales back down when the session expires — with one difference: **its readiness and errors never gate the session**. The waiting page clears (and a blocking request returns) as soon as all non-optional instances are ready, even if an optional instance is still starting, unhealthy, or failed to start entirely.
+
+Use this to split a stack into important instances that visitors must wait for and auxiliary ones (workers, reporting, developer tooling) that should come up alongside them but must never block the stack when they are broken.
+
+## How it works
+
+The `optional:` marker travels inside the instance name, so every reverse proxy plugin forwards it unchanged — no plugin upgrade is needed. Sablier strips the prefix when the session is requested; providers, sessions and metrics all see the clean instance name. No workload name can collide with the marker because Docker, Swarm, Podman and Kubernetes names cannot contain a colon.
+
+The prefix works the same in a direct API call:
+
+```text
+GET /api/strategies/dynamic?names=whoami&names=optional:whoami-helper&session_duration=5m
+```
+
+## Notes
+
+- A session whose members are **all** optional is immediately ready.
+- Optional instances still appear on the waiting page details while it is shown, but they are not listed as a blocking reason in timeout errors.
+- Optionality is per request: the same instance can be required in one middleware and optional in another.
+- For group-based sessions, instances are discovered via labels and cannot be marked optional yet; use `names` for split behaviour.
+
+## Related
+
+- [Mark ready on start](/how-to-guides/readiness/ready-on-start/) — skip the health check but still require a successful start.
+- [Settling delay](/how-to-guides/readiness/ready-after/) — the opposite direction: require extra readiness time.
--- a/docs/static/openapi.json
+++ b/docs/static/openapi.json
@@ -8,6 +8,9 @@
           },
           "instance": {
             "$ref": "#/components/schemas/sablier.InstanceInfo"
+          },
+          "optional": {
+            "type": "boolean"
           }
         },
         "type": "object"
@@ -528,7 +531,7 @@
         "description": "Holds the request until the requested instances are ready, or until the timeout elapses. Provide either `names` (one or more) or `group`, never both.",
         "parameters": [
           {
-            "description": "Instance name(s); repeat for multiple. Mutually exclusive with group.",
+            "description": "Instance name(s); repeat for multiple. Mutually exclusive with group. Prefix a name with 'optional:' to mark it best-effort: it is started with the session but its readiness and errors never gate it.",
             "in": "query",
             "name": "names",
             "schema": {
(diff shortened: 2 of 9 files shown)


acouvreur:
Hello @thisispr 

Thank you for your contribution. I would rather avoid changing anything to the "names" query parameter API.

I don't like it, and I would prefer to use the group discovery mechanism instead.

---

What do you think ?

I'm not closed to adding this feature, but I think it would be best suited on the instance label directly ?

```
- sablier.optional
```

I think it's also the same thing that @ngoc-minh-do  wanted to do in https://github.com/sablierapp/sablier/pull/999

Basically a pass-through / poke strategy.

thisispr:
> Hello @thisispr
> 
> Thank you for your contribution. I would rather avoid changing anything to the "names" query parameter API.
> 
> I don't like it, and I would prefer to use the group discovery mechanism instead.
> 
> What do you think ?
> 
> I'm not closed to adding this feature, but I think it would be best suited on the instance label directly ?
> 
> ```
> - sablier.optional
> ```
> 
> I think it's also the same thing that @ngoc-minh-do wanted to do in #999
> 
> Basically a pass-through / poke strategy.

Hi @acouvreur, thanks for the quick feedback!


Funny enough, we went down the label road first — this PR is the result of what we learned there. Context: we run Sablier on Kubernetes (Traefik plugin, `names`-based middlewares) to scale whole dev environments (~80 deployments each) to zero, and we prepared the label-based split (using `sablier.ready-on-start` as the closest existing thing) across all our Helm values before abandoning it. Two concrete findings:

**1. A label can't reach every workload.** Our own charts let us inject Deployment-metadata labels, but several third-party charts (temporal, superset, community workflow workers) expose no hook for Deployment *metadata* labels at all — and those auxiliary services are exactly the ones that most need to be optional. The middleware's `names` list is the one place that already enumerates every managed instance, regardless of who authored its chart.

**2. A label is static per workload, but optionality is a property of the route.** This is the same point as @ngoc-minh-do's Frigate example in #999: our reporting UI should be *optional* when the whole environment wakes through the main application's middleware, but *gating* (show the waiting page) when a user opens the reporting URL directly through its own middleware. A `sablier.optional` label can only express one of the two; a request-scoped marker expresses both:

```mermaid
flowchart LR
    subgraph A["sablier.optional label (static per workload)"]
        direction LR
        L1["own charts&nbsp;&nbsp;✅ label possible"] --> D1[Deployment metadata]
        L2["3rd-party charts&nbsp;&nbsp;❌ no label hook"] -.-> D1
        D1 --> S1[Sablier]
        S1 --> B1["same behaviour on **every** route"]
    end
```

```mermaid
flowchart LR
    subgraph B["request-scoped optional: (this PR)"]
        direction LR
        M1["middleware <i>environment</i><br/>names: app, <b>optional:</b>reporting"] --> S2[Sablier]
        M2["middleware <i>reporting</i><br/>names: reporting"] --> S2
        S2 --> B2["reporting: background-start on env wake,<br/>waiting page on direct access"]
    end
```

That said, I completely understand not wanting to overload the `names` values — it is a bit of a smuggled parameter. Ranked by our preference, we'd gladly rework toward any of these:

1. **Keep the request-scoped marker (this PR)** — works today with every released plugin, zero plugin changes.
2. **Do both**: `sablier.optional` label for group discovery (nice default for label-centric setups) *plus* the request-scoped marker for `names` mode. The label sets the default, the request can override. Happy to add the label part to this PR.
3. **A dedicated `optional_names` query parameter** — the cleanest API, but every proxy plugin needs a release before anyone can use it, and `names`/`optional_names` must be kept consistent by the user.

If you feel strongly about label-only we can contribute that instead — it would still be a big improvement — but it wouldn't cover the third-party-chart and per-route cases above. What do you think?

ngoc-minh-do:
@thisispr #999 was closed in favor of #1015, which adds a poke strategy: start instances, don't wait, pass the request through. It's per-route (middleware), not per-workload label. Do you think #1015 could help in your situation?

thisispr:
@ngoc-minh-do thanks for the pointer — I had missed #1015, and yes, I think it covers our case.

Chaining two middlewares on the main router would give exactly the important/auxiliary split we're after:

```yaml
# router: middlewares: [env-important, env-auxiliary]
env-important:            # gates the waiting page
  plugin:
    sablier:
      names: app-frontend, app-api, ...
      dynamic: {}
env-auxiliary:            # poked: started, never gates
  plugin:
    sablier:
      names: reporting, workers, ...
      poke: {}
```

It also addresses both concerns from my comment above: the poke middleware is names/group-scoped, so it reaches third-party workloads we cannot label, and it stays per-route — a dedicated `dynamic` middleware on the reporting URL still shows the waiting page on direct access. And it has a bonus this PR doesn't offer: the auxiliary set gets its own `sessionDuration`.

Two practical notes from our side:

- #1015 only becomes usable once the companion plugin PRs are released; the marker in this PR works with every already-released plugin version (we pin sablier-traefik-plugin v1.3.0). That's the main remaining difference between the two approaches.
- Two middlewares mean two sessions and one extra Sablier round-trip per request — acceptable for us.

@acouvreur if you prefer the poke direction, we're happy to close this PR in favor of #1015 — and to help it land: we can test it against our ~80-deployment dev environments and review the sablier-traefik-plugin companion PR. Just say the word.

thisispr:
@acouvreur Do you have time to take a look at this? Is there anything I can do to help?

acouvreur (approved):
So far this has a very small radius blast, and the feature is only available to named entries.

We can merge it if everything is green.

PR #1091, opened 2026-08-24
docs: add a Pangolin reverse proxy guide

### LINKED ISSUE #1040: Add Pangolin reverse proxy integration
**Describe the reverse proxy you'd like**
https://github.com/fosrl/pangolin

**Does the reverse proxy provide middleware/module/extension support**
Pangolin uses Traefik
For traefik it's http middleware, for nginx it's modules etc.

acouvreur:
Closes #1040

Pangolin isn't a reverse proxy of its own — it drives a Traefik v3 instance whose dynamic configuration it generates from its own database. So it integrates through the existing Traefik plugin, and there's nothing Pangolin-specific to install on the Sablier side.

The guide therefore covers only what *is* Pangolin-specific:

- **Loading the plugin** next to Pangolin's own `badger` in `config/traefik/traefik_config.yml`, which is a user-editable file in every install.
- **Making Sablier reachable from Traefik.** This is the part #1040 is actually about: Sablier needs the Docker socket so it runs at the site, while Traefik runs on the VPS. Covers reaching it across a Newt tunnel, and the `DOCKER_HOST` alternative.
- **Attaching the middleware to a resource.** Pangolin builds every router's chain as `[badger, ...traefik.additional_middlewares]` and has no per-resource middleware field, so there are two approaches with different trade-offs — global via `additional_middlewares`, or per-resource by shadowing the generated router in the file provider.

Adds the page under `tutorials/reverse-proxies/`, a card on the index, and the Pangolin logo.

## Verified

Everything here was checked against a running Pangolin stack rather than written from the docs. The generated router comes back as:

```json
{
  "name": "1-mimic-router@http",
  "rule": "Host(`mimic.localhost`)",
  "middlewares": ["badger@http", "sablier-mimic@file"],
  "service": "1-mimic-service",
  "priority": 100
}
```

Both attachment options were confirmed working end to end, including the cross-provider `@http` service reference used by the per-resource option. `hugo` builds clean with no shortcode leakage.

The runnable stack behind that verification is a companion PR: sablierapp/sablier-traefik-plugin#61

## Note on the ordering claim

The guide states that `badger` runs before Sablier, so the waiting page is only ever served to authenticated users. That's not incidental — it falls out of Pangolin always prepending `badger` to the chain, which is why the ordering is safe by construction rather than by configuration.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


acouvreur:
Pushed an update adding [Middleware Manager](https://github.com/hhftechnology/middleware-manager) as an attachment option, and **correcting a claim in the original PR**.

Middleware Manager is a community service listed in [Pangolin's own docs](https://docs.pangolin.net/self-host/community-guides/middlewaremanager). It reads resources from Pangolin's internal API and re-serves Traefik's dynamic configuration with middlewares attached per resource — solving the per-resource gap the guide describes, with discovery and a UI instead of hand-maintained routers.

## The correction

The guide previously stated that `badger` runs before Sablier, so the waiting page is only ever served to authenticated users. That holds for `additional_middlewares` and for a hand-written shadow router, but **not** for Middleware Manager, which unconditionally places its own additions *before* the router's existing middlewares. The chain comes out as:

```
["sablier-mimic@file", "badger@http"]
```

`priority` only orders Middleware Manager's own assignments among themselves — I verified at priorities `1` and `-100` that it cannot move one after `badger`.

So an unauthenticated request can wake the container. On a zero-trust proxy that's a real consideration, and it means none of the three options wins outright — hence a trade-off table rather than a recommendation:

| | Per-resource | Auth before wake | Extra service |
| --- | --- | --- | --- |
| `additional_middlewares` | ✗ global | ✓ | ✗ |
| Middleware Manager | ✓ | ✗ | ✓ |
| Shadow the router | ✓ | ✓ | ✗ |

Every claim here was checked against the running stack in the companion PR, not read off documentation. `hugo` builds clean and all three in-page anchors resolve.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

